PR #8, opened 2020-02-20
reinstate kessler_update which was accidentally removed

### LINKED ISSUE #6: Missing kessler_update?
suite_kessler.xml still has a call to kessler_update but that file does not seem to have made it to the current master. Did it get lost in the directory rearrangement?

cacraigucar:
reinstate kessler_update.  Should be an exact copy of what was there before.

Closes #6

### DIFF
--- a/kessler/kessler_update.F90
+++ b/kessler/kessler_update.F90
@@ -66,16 +66,15 @@ CONTAINS
       integer,            intent(out)   :: errflg
 
       integer                           :: klev
-      real(kind_phys)                   :: ptend_s(ncol)
+      real(kind_phys)                   :: new_temp(ncol)
 
       errmsg = ''
       errflg = 0
 
       ! Back out tendencies from updated fields
       do klev = 1, nz
-         ptend_s(:ncol) = ((theta(:ncol,klev) * exner(:ncol,klev)) -          &
-              temp_prev(:ncol,klev)) * cpair / dt
-         ttend_t(:ncol,klev) = ttend_t(:ncol,klev) + (ptend_s(:ncol) / cpair)
+         new_temp(:ncol) = theta(:ncol,klev) * exner(:ncol,klev)
+         ttend_t(:ncol,klev) = ttend_t(:ncol,klev) + ((new_temp(:ncol) - temp_prev(:ncol,klev)) / dt)
       end do
 
    end subroutine kessler_update_run
--- a/kessler/kessler_update.F90
+++ b/kessler/kessler_update.F90
@@ -78,12 +78,6 @@ CONTAINS
          ttend_t(:ncol,klev) = ttend_t(:ncol,klev) + (ptend_s(:ncol) / cpair)
       end do
 
-      ! Save the temperature for the next time step
-      !!XXgoldyXX ==> @cacraigucar: Does this have any effect?
-      temp_prev(:,:)     = temp(:,:)
-
-!    surf_state%precl(:ncol) = surf_state%precl(:ncol) + precl(:ncol)  ! KEEPING THIS HERE AS A REMINDER
-
    end subroutine kessler_update_run
 
    !> \section arg_table_kessler_update_timestep_final  Argument Table
--- a/kessler/kessler_update.F90
+++ b/kessler/kessler_update.F90
@@ -1,0 +1,118 @@
+module kessler_update
+
+   use ccpp_kinds, only: kind_phys
+
+   implicit none
+   private
+
+   public :: kessler_update_init
+   public :: kessler_update_timestep_init
+   public :: kessler_update_run
+   public :: kessler_update_timestep_final
+
+   ! Private module variables
+   real(kind_phys)    :: gravit
+   real(kind_phys)    :: cpair
+
+CONTAINS
+
+   !> \section arg_table_kessler_update_init  Argument Table
+   !! \htmlinclude kessler_update_init.html
+   subroutine kessler_update_init(gravit_in, cpair_in)
+      real(kind_phys),    intent(in)    :: gravit_in
+      real(kind_phys),    intent(in)    :: cpair_in
+
+      gravit = gravit_in
+      cpair = cpair_in
+
+   end subroutine kessler_update_init
+
+   !> \section arg_table_kessler_update_timestep_init  Argument Table
+   !! \htmlinclude kessler_update_timestep_init.html
+   subroutine kessler_update_timestep_init(temp, temp_prev, ttend_t,          &
+        errmsg, errflg)
+
+      real(kind_phys),    intent(in)  :: temp(:,:)
+      real(kind_phys),    intent(out) :: temp_prev(:,:)
+      real(kind_phys),    intent(out) :: ttend_t(:,:)
+      character(len=512), intent(out) :: errmsg
+      integer,            intent(out) :: errflg
+
+      !   Initialize the previous temperature and its tendency to zero
+      temp_prev(:,:)  = temp(:,:)
+      ttend_t(:,:)    = 0._kind_phys
+
+      errmsg = ''
+      errflg = 0
+
+   end subroutine kessler_update_timestep_init
+
+   !> \section arg_table_kessler_update_run  Argument Table
+   !! \htmlinclude kessler_update_run.html
+   subroutine kessler_update_run(nz, ncol, temp, theta, exner, dt,      &
+        temp_prev, ttend_t, errmsg, errflg)
+
+      integer,            intent(in)    :: nz
+      integer,            intent(in)    :: ncol
+      real(kind_phys),    intent(in)    :: temp(:,:) ! temperature
+      real(kind_phys),    intent(in)    :: exner(:,:)
+      real(kind_phys),    intent(in)    :: dt
+
+      real(kind_phys),    intent(inout) :: theta(:,:)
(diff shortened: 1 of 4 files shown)


kessler/kessler_update.meta
```diff
@@ -0,0 +1,194 @@
...
+[ temp ]
+  standard_name = temperature
+  state_variable = true
+  type = real
+  kind = kind_phys
+  units = K
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  intent = in
+[ zm ]
+  standard_name = geopotential
```
nusbaume: I suspect this is geopotential height, not geopotential (as the units are in meters)

kessler/kessler_update.F90
```diff
@@ -0,0 +1,118 @@
...
+      ! Back out tendencies from updated fields
+      do klev = 1, nz
+         ptend_s(:ncol) = ((theta(:ncol,klev) * exner(:ncol,klev)) -          &
+              temp_prev(:ncol,klev)) * cpair / dt
+         ttend_t(:ncol,klev) = ttend_t(:ncol,klev) + (ptend_s(:ncol) / cpair)
+      end do
+
+      ! Save the temperature for the next time step
+      !!XXgoldyXX ==> @cacraigucar: Does this have any effect?
+      temp_prev(:,:)     = temp(:,:)
```
nusbaume: It looks like I am just repeating @gold2718's comment here, but is this line needed if we are already setting `temp_prev` in the `kessler_update_timestep_init` subroutine?

kessler/kessler_update.meta
nusbaume: I suspect this is geopotential height, not geopotential (as the units are in meters).  I realize that in the kessler.meta file we also label it "geopotential", but given that geopotential and geopotential height are two physically different quantities, it might be good to have both of them labeled  "geopotential height", at least for their standard names.  

I'll go ahead and create an issue for the kessler.meta file standard name.

kessler/kessler_update.F90
```diff
@@ -0,0 +1,118 @@
...
+
+      integer                           :: klev
+      real(kind_phys)                   :: ptend_s(ncol)
+
+      errmsg = ''
+      errflg = 0
+
+      ! Back out tendencies from updated fields
+      do klev = 1, nz
+         ptend_s(:ncol) = ((theta(:ncol,klev) * exner(:ncol,klev)) -          &
```
nusbaume: I realize that we don't really want to change the physics calculations themselves during the switch to the CCPP framework, but if `ttend_t` is the only output, then one could calculate that directly without `ptend_s` like so:

`ttend_t(:ncol,klev) = ttend_t(:ncol,klev) + ((theta(:ncol,klev) * exner(:ncol,klev)) - temp_prev(:ncol,klev)) / dt`

Otherwise this routine is needlessly converting temperature to thermal energy, and then immediately converting that thermal energy back into temperature.

gold2718: Agreed, this looks like a leftover from the current CAM `ptend` vs. `tend` split where `ptend` calculated a heating rate while `tend` stored a temperature tendency. Of course, this gets me thinking that maybe the standard_name for `ttend_t` is not quite right since it may not be a "total" temperature tendency all by itself.
I see no reason to avoid this change. I do not think that kessler_update will called from CAM.

kessler/kessler_update.meta
gold2718: Agreed, this is a bug due to a rushed reading of the names GFS physics is using. Will comment on #11 but just wanted to mention that the new standard names library is at https://github.com/ESCOMP/CCPPStandardNames.

kessler/kessler_update.F90
cacraigucar: Removed

kessler/kessler_update.meta
cacraigucar: Renamed - Note that this will require a change to CAMDEN as I needed to change the name in the registry.xml also.

kessler/kessler_update.F90
cacraigucar: I decided to not change this.  At this point, I'm trying to maintain the "original code" as much as possible for traceability.  Feel free to point out that you disagree that this is not the way to approach the transition.

gold2718: @cacraigucar, @nusbaume: Please do not "Resolve conversation" unless you are the conversation starter. This makes sure that the issue is resolved to the satisfaction of the reviewer before hiding it from view.
Responding to your point, while I agree in principle that we should not make changes for the fun of it, this conversion process represents an unprecedented opportunity to clean up our code. Imagine a graduate student, years from now, getting confused by why the code is this way, or worse, using it as an example of 'good parameterization coding style' and propagating it. After all, the simpler physics code is what is taught in school.
On the third hand, this very particular case is likely to go away when we no longer need to manually back out tendencies.
On the fourth hand, when that happens, it would be nice to obtain the same answer as the current manual process.

cacraigucar: So which hand wins?  

Also, the reason I "resolve conversation" is that on big commits (like the cam_snapshot), I totally get lost with what I had completed and what was still outstanding.  I got into the habit of "resolving conversations" when I had addressed it.  BTW, I'm not sure if all reviewers will go back and "resolve conversations".  I guess I can add my own special keyword if you are really against my workflow.

gold2718: I think I am with 'The Fourth Hand' (sounds like some sort of thriller or horror movie).

We should discuss the workflow, however, just because you (or any of us) have 'addressed' an issue does not mean is has been 'completed'. Once you click 'Resolve conversation', it then becomes even harder for the reviewer to understand what to focus on at re-review time.

How about adding a entry to the conversation specifying what you did to address the issue? Then, any conversation where you (currently) have the last word is something you do not have to worry about until the re-reviews come in.

kessler/kessler_update.meta
```diff
@@ -0,0 +1,194 @@
...
+  intent = in
+[ zm ]
+  standard_name = geopotential_height
+  state_variable = true
+  type = real | kind = kind_phys
+  units = m
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  intent = in
+[ phis ]
+  standard_name = geopotential_height_at_surface
```
nusbaume: Sorry to be confusing, but the variable `phis` actually is "geopotential", so the standard name should say:

`geopotential_at_surface`

Basically if the units are `m` then it is geopotential height, whereas if the units are `m2 s-2` then it is geopotential.

kessler/kessler_update.F90
cacraigucar: OK - I'll just ignore the resolve conversation button.  

Drat - you chose the fourth hand and I'm not sure what you mean by it.   If you are requesting a change, let me know.  I assume it would be to implement Jesse's suggestion.  If you are not requesting a change, then I believe I'm done.

kessler/kessler_update.meta
cacraigucar: Resolved

kessler/kessler_update.F90
gold2718: See, I told you "The Fourth Hand" was a horror movie waiting to be made ;)
In this case, I think it means replacing lines 76 - 78 with:
```
new_temp(:ncol) = theta(:ncol,klev) * exner(:ncol,klev)
ttend_t(:ncol,klev) = ttend_t(:ncol,klev) + ((new_temp(:ncol) - temp_prev(:ncol,klev)) / dt)
```
where `new_temp` (or some similar name) replaces `ptend_s` as a local variable.

kessler/kessler_update.meta
nusbaume: Thanks!

kessler/kessler_update.F90
nusbaume: I agree that it would be good to get rid of the unnecessary `cpair` calculations.  I am even fine with keeping the original code in, but commented out (which I realize violates all of our coding standards), if there are concerns about traceability, simply because as currently written we will be wasting cycles doing division for no reason.

Of course, if there is a concern that this modification will cause meaningful answer changes (which it shouldn't, but you never know) then I am fine to leave it as-is for now.

cacraigucar: Put in change as described by @gold2718.  For the 9 timestep case it did not change answers.  This will need to be revisited if answers don't agree on longer runs.

cacraigucar: Resolved

nusbaume: Thanks!  Once these changes (and the `phis` changes below) have been added to this PR I'll go ahead and approve it.

gold2718:
Do you always use the 'full' version of `push`?
```
git push <remote_name> <branch_name>
```
To avoid doing something accidentally, set this configuration on all your machines:
```
git config --global push.default nothing
```

cacraigucar:
I think I know some of what I did wrong...

Yes, I always say 'git push origin my_branch_name'.  The problem is that origin was pointing to NCAR instead of my fork.

I picked up the project after having it sit idle for a few weeks.  For whatever reason, right before I left the project I must have pulled the directory straight from the NCAR repo.   When I picked up the project, I saw it had a detached head.  After verifying that the hash was correct, I switched/made the branch and made my changes.  I then pushed the changes back to origin (which was NCAR).

I just need to make a point of doing a git remote -v before I do a git push, to make sure I'm pushing to the location I want.

PR #16, opened 2020-06-02
Fix safe set

### LINKED ISSUE #15: Wrong intent in safe_set
The `safe_set` routine in state_converters.F90 has `var` as `intent(out)`. However, this allows the value of `var` to be undefined as you enter the routine. When this happens (happens sometimes with nag compiler), the test for `(var == unset)` will fail. Also, the `else` clause fails to set the value which is an error.
I suggest that the intent be change to `inout`.

cacraigucar:
NAG debug uncovers a problem with variable "var" in safe_set.  This variable needs to be intent(inout)

Closes #15


cacraigucar:
Wait a sec...

My Held-Suarez changes are in here too and I don't want them yet.

cacraigucar:
OK - my two letter commit is ready for review!

nusbaume:
Actually, Do we want to reference issue #15 above so that it is closed once this PR is merged?

PR #18, opened 2020-06-18
Kessler physics computes a consistent relative humidity

### LINKED ISSUE #17: Kessler physics needs to compute relative humidity
Kessler needs to produce the relative humidity instead of having it computed in cam_diagnostics. This guarantees that the maximum relative humidity is 100%.

cjablono:
Instead of computing the relative humidity in cam_diagnostics, the relative humidity is now computed in the Kessler routine. This is now consistent with the computation of the vapor saturation value, and produces a maximum relative humidity of 100%.
Fixes #17

### DIFF
--- a/kessler/kessler.F90
+++ b/kessler/kessler.F90
@@ -127,7 +127,7 @@ CONTAINS
    !> \section arg_table_kessler_run  Argument Table
    !! \htmlinclude kessler_run.html
    subroutine kessler_run(ncol, nz, dt,  lyr_surf, lyr_toa, rho, z, pk, theta, &
-        qv, qc, qr, precl, errmsg, errflg)
+        qv, qc, qr, precl, relhum, errmsg, errflg)
 
       !------------------------------------------------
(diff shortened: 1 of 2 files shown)


gold2718:
Yes, the CESM2.2 release is the priority, however, the PR branch _should_ also work as-is in CAMDEN. Let me know if it does not.

cacraigucar:
For a 9 timestep FKESSLER run using the cam_development branch, I am seeing identical answers to the prior version.  I assume that this is possible.  @gold2718 What testing do we want to make to validate this change in CAM?

gold2718:
@cjablono, any idea what would be a good test to make sure we have the fix plugged in correctly? Is there a test case that exercises the issue and would show a difference?

gold2718:
@brian-eaton, can you have a look at this so we can get it in and tag it? We need the updated tag to finish the Simpler Models PR.

cacraigucar:
This has now been tagged version0_0_006

PR #23, opened 2020-11-06
Bring in CCPP'ized Held Suarez

### LINKED ISSUE #14: Implement Held-Suarez suite
Port the Held-Suarez suite to the CCPP.

cacraigucar:
Add the CCPP metadata for Held Suarez and minor changes to get it to work in CCPP.

Tests run:
Brought into cam6_3_002 and the post "snapshot" agrees with the CAM repo's post snapshot
Brought into cam6_3_002 and the izumi PGI and NAG tests all pass (except for the DAE known failure)

Ran test using CAMDEN:
./create_newcase --case test_held_suarez_try28 --compset FPHYStest --res f19_f19 --compiler intel --run-unsupported --pecount 1
 ./xmlchange DOUT_S=FALSE ; ./xmlchange CAM_CONFIG_OPTS="--dyn none --physics-suites held_suarez"; ./xmlchange NTASKS=1 ;  ./xmlchange DEBUG=TRUE;  ./xmlchange STOP_N=9; ./xmlchange STOP_OPTION=nsteps ; cp ../user_nl_cam .

Contents of user_nl_cam:
 ncdata         = '/home/cacraig/held_suarez_data/run_heldsuarez_try11.cam.h5.0001-01-01-00000.nc' 
pertlim        =  1.0D-5 
use_topo_file  =  .false.

Closes #14


held_suarez/held_suarez.F90
```diff
@@ -0,0 +1,166 @@
...
+    !
+    ! Compute idealized radiative heating rates (as dry static energy)
+    !
+    !
+    do k = 1, pver
+      if (pref_mid_norm(k) > sigmab) then
+        do i = 1, ncol
+          kt = ka + (ks - ka)*cossqsq(i)*(pref_mid_norm(k) - sigmab)/onemsig
+          trefc   = 315._kind_phys - (60._kind_phys * sinsq(i))
+          trefa = (trefc - 10._kind_phys*cossq(i)*log((pmid(i,k)/psurf_ref)))*(pmid(i,k)/psurf_ref)**cappa
```
nusbaume, line 134: I would convert the `10._kind_phys` number into its own parameter, as it is representing a specific, potentially tunable, physical quantity (the change in potential temperature over one scale height of the atmosphere), at least according to the original manuscript.

held_suarez/held_suarez.F90
```diff
@@ -0,0 +1,166 @@
...
+    real(kind_phys), intent(in)  :: pmid(:,:) ! mid-point pressure
+    real(kind_phys), intent(in)  :: u(:,:)    ! Zonal wind (m/s)
+    real(kind_phys), intent(in)  :: v(:,:)    ! Meridional wind (m/s)
+    real(kind_phys), intent(in)  :: t(:,:)    ! Temperature (K)
+                                              !
+                                              ! Output arguments
+                                              !
+    real(kind_phys), intent(out) :: du(:,:)   ! Zonal wind tend
+    real(kind_phys), intent(out) :: dv(:,:)   ! Meridional wind tend
+    real(kind_phys), intent(out) :: s(:,:)    ! Heating rate
```
nusbaume, line 89: Could we possibly rename the variable `s` to `ds`, to indicate that it is also a tendency (usually `s` by itself represents the dry static energy, not the heating rate).

held_suarez/held_suarez.meta
```diff
@@ -0,0 +1,85 @@
...
+  intent = in
+[ psurf_ref_in ]
+  standard_name = reference_pressure_at_sea_level
+  units = Pa
+  type = real | kind = kind_phys
+  dimensions = ()
+  intent = in
+[ pref_mid_norm_in ]
+  standard_name = reference_pressure
+  units = Pa
```
nusbaume, line 24: I believe the units here should be `1`, as the pressure is normalized.

held_suarez/held_suarez.meta
```diff
@@ -0,0 +1,85 @@
...
+  intent = out
+[ dv ]
+  standard_name = tendency_of_northward_wind
+  units = m s-2
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent,vertical_layer_dimension)
+  intent = out
+[ s ]
+  standard_name = tendency_of_temperature
+  units = K s-1
```
nusbaume, line 82: It might make more sense for this variable to be labeled `ds`, instead of `s`, to clarify that it is in fact a tendency, not a state variable.

Also, the standard name here should say `tendency_of_dry_static_energy`, or something equivalent, as the temperature is being multiplied by `cpair` in the actual fortran routine (which converts it to static energy).

Finally, the units here should be `J kg-1 s-1`.

nusbaume (changes requested):
It's looking good, but I did find two variables that had the wrong units, and I have a few other requests which should hopefully make the physics scheme easier to understand (at least for other folks with an atmospheric science background).

PR #38, opened 2021-09-27
qneg module parameterization

peverwhee:
QNEG parameterization: 
- Added utilities/qneg.F90 and .meta
- Added qneg to existing physics scheme SDFs
- Added scheme_name output variable in existing physics schemes (kessler, held suarez) - to be input to qneg

This has been tested with KESSLER with the null dycore and the SE dycore. Tested by:
- Manually setting qmin to 1e-3 for rain water mixing ratio
- Comparing the number of violations (and the worst violation) in new parameterization in CAMDEN to the same metrics in CAM [when called from DP_COUPLER and KESSLER]


utilities/qneg.F90
```diff
@@ -0,0 +1,509 @@
...
+  interface reset_stats
+     module procedure reset_stats_scalar
+     module procedure reset_stats_array
+  end interface reset_stats
+
+contains
+
+!> \section arg_table_qneg_init Argument Table
+!! \htmlinclude qneg_init.html
+  subroutine qneg_init(print_qneg_warn, number_of_constituents, qmin)
```
gold2718, line 62: Should have `errcode` and `errmsg` (initialize them to zero and empty string respectively).

utilities/qneg.F90
```diff
@@ -0,0 +1,509 @@
...
+       case('timestep')
+          collect_stats = .true.
+          timestep_reset = .true.
+       case('off')
+          collect_stats = .false.
+          timestep_reset =.false.
+       case default
+          !do nothing
+!          call endrun(sub//"FATAL: '"//trim(print_qneg_warn)//"' is not a    &
+!             valid value for print_qneg_warn")
```
gold2718, line 117: Set `errcode = 1` and `errmsg` to this message.

utilities/qneg.F90
```diff
@@ -0,0 +1,509 @@
...
+!          index = i
+!          call reset_stats(qneg4_warn_num(index), qneg4_warn_worst(index))
+!          exit
+!       end if
+!    end do
+!  end function find_index4
+
+!> \section arg_table_qneg_run Argument Table
+!! \htmlinclude qneg_run.html
+  subroutine qneg_run (subnam, ncol, ncold, lver, num_constituents, qmin, q)
```
gold2718, line 185: Should have `errcode` and `errmsg` (initialize them to zero and empty string respectively).

gold2718 (changes requested):
A few changes needed.

kessler/kessler.F90
```diff
@@ -21,14 +21,15 @@ module kessler
 
    !> \section arg_table_kessler_init  Argument Table
    !! \htmlinclude kessler_init.html
-   subroutine kessler_init(cp_in, lv_in, psl_in, rair_in, rhoqr_in, errmsg, errflg)
+   subroutine kessler_init(cp_in, lv_in, psl_in, rair_in, rhoqr_in, scheme_name, errmsg, errflg)
```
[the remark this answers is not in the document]
peverwhee: moved to run routine.

held_suarez/held_suarez_1994.F90
```diff
@@ -52,13 +52,14 @@ module held_suarez_1994
 
 !> \section arg_table_held_suarez_1994_init Argument Table
 !! \htmlinclude held_suarez_1994_init.html
-  subroutine held_suarez_1994_init(pver_in, cappa_in, cpair_in, psurf_ref_in, pref_mid_norm_in, errmsg, errflg)
+  subroutine held_suarez_1994_init(pver_in, cappa_in, cpair_in, psurf_ref_in, pref_mid_norm_in, scheme_name, errmsg, errflg)
```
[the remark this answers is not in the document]
peverwhee: also moved to run routine

utilities/qneg.meta
```diff
@@ -0,0 +1,116 @@
...
+  units = count
+  type = integer
+  dimensions = ()
+  intent = in
+[ ncold ]
+  standard_name = horizontal_dimension
+  units = count
+  type = integer
+  dimensions = ()
+  intent = in
```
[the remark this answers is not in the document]
peverwhee: changed to horizontal_loop_begin/end

utilities/qneg.meta
```diff
@@ -0,0 +1,116 @@
...
+  units = 1
+  type = character | kind = len=*
+  dimensions = ()
+  intent = in
+[ ncol ]
+  standard_name = horizontal_loop_extent
+  units = count
+  type = integer
+  dimensions = ()
+  intent = in
```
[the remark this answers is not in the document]
peverwhee: changed to horizontal_loop_begin/end

utilities/qneg.F90
```diff
@@ -0,0 +1,509 @@
...
+  private
+  save
+
+  ! Public interface.
+
+  public :: qneg_init
+  public :: qneg_run !qneg3 in CAM6
+  public :: qneg_timestep_final
+  public :: qneg_final
+!  public :: qneg4
```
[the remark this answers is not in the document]
peverwhee: removed qneg4 & all 3 suffixes

utilities/qneg.F90
peverwhee: good call - added in both routines

utilities/qneg.F90
```diff
@@ -0,0 +1,371 @@
...
+!          call outfld(trim(diag_names(num_constituents+m)), badcols, pcols, idx)
+!       end if
+    end do
+    call t_stopf ('qneg_run')
+
+  end subroutine qneg_run
+
+!> \section arg_table_qneg_timestep_final Argument Table
+!! \htmlinclude qneg_timestep_final.html
+  subroutine qneg_timestep_final(mpi_communicator, rootprocid, isrootproc)
```
gold2718, line 281: Also needs `errcode` and `errmsg` (sorry for not noticing before).

utilities/qneg.F90
```diff
@@ -0,0 +1,371 @@
...
+    logical, intent(in) :: isrootproc
+
+    if (timestep_reset .and. collect_stats) then
+       call qneg_print_summary(mpi_communicator, rootprocid, isrootproc)
+    end if
+
+  end subroutine qneg_timestep_final
+!> \section arg_table_qneg_final Argument Table
+!! \htmlinclude qneg_final.html
+  subroutine qneg_final(mpi_communicator, rootprocid, isrootproc)
```
gold2718, line 294: Also needs `errcode` and `errmsg` (sorry for not noticing before).

gold2718 (approved):
Looking good. Please add the two interfaces I did not notice the first time.

suite_cam6.xml
```diff
@@ -242,7 +242,7 @@
       <!-- **************************** -->
       <!-- Negative moisture/tracer check -->
       <!-- **************************** -->
-      <scheme>qneg4</scheme>                          <!-- Modify surface moisture and heat fluxes to prevent negative moisture values, and output diagnostics -->
+<!--      <scheme>qneg4</scheme>                         --><!-- Modify surface moisture and heat fluxes to prevent negative moisture values, and output diagnostics -->
```
nusbaume, line 245: I might leave this uncommented, but rename `qneg4` to `qneg_surface`, with a comment noting that in CAM this is referred to as "qneg4".  The reason for this change is because every uncommented scheme in this file indicates a physics or chemistry process that must exist in some form as a CCPP scheme in order to exactly replicate CAM6 physics, and my fear is that if we comment it out my future self might think that it is un-necessary.

utilities/qneg.F90
```diff
@@ -0,0 +1,381 @@
...
+!    call addfld(diag_names((2*num_constituents) + 1), horiz_only, 'I', 'kg/m^2/s',     &
+!         'qflux excess (QNEG4)')
+
+    !Create fake qmin -> will be removed when qmin exists
+    qmin = 0._kind_phys
+
+    !Allocate and initialize arrays whose dimensions depend on num_constituents:
+    num_constituents = number_of_constituents
+    allocate(qneg_warn_num(num_constituents, num_bins), stat=ierr)
+    call check_allocate(ierr, subname, 'qneg_warn_num')
```
nusbaume, line 97: It might be worth discussing at some point how CCPP schemes can raise an error associated with allocation failures without having to bring-in the `check_allocate` function via a CAMDEN-specific `use` statement.  Maybe add this to the issue I mentioned above as a discussion topic?

suite_cam6.xml
peverwhee: I'm on board - changed to qneg_surface and uncommented.

utilities/qneg.F90
gold2718: Oh, good point. I am so used to seeing the `check_allocate` calls that I forgot we were in a CCPP scheme. These should be using `errcode` and `errmsg` followed by `return`.

suite_cam6.xml
gold2718: I think that `qneg_surface` should be its own scheme (in its own module). I agree that it should be uncommented here but maybe open an issue to remind us that it needs to be implemented.

peverwhee: issue opened.

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
...
+    allocate(qneg_warn_num(num_constituents, num_bins), stat=ierr)
+    if (ierr /= 0) then
+       errcode = 1
+       errmsg = trim(subname)//': Allocate of qneg_warn_num failed'
+       return
+    end if
+    allocate(qneg_warn_worst(num_constituents, num_bins), stat=ierr)
+    if (ierr /= 0) then
+       errcode = 1
+       errmsg = trim(subname)//': Allocate of qneg_warn_worst'
```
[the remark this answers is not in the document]
peverwhee: indeed!

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
...
+!         'qflux excess (QNEG4)')
+
+    !Create fake qmin -> will be removed when qmin exists
+    qmin = 0._kind_phys
+
+    !Allocate and initialize arrays whose dimensions depend on num_constituents:
+    num_constituents = number_of_constituents
+    allocate(qneg_warn_num(num_constituents, num_bins), stat=ierr)
+    if (ierr /= 0) then
+       errcode = 1
```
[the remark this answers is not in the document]
peverwhee: changed.

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
...
+!    end do
+!    diag_names((2*num_constituents) + 1) = 'qflux_exceeded'
+!    call addfld(diag_names((2*num_constituents) + 1), horiz_only, 'I', 'kg/m^2/s',     &
+!         'qflux excess (QNEG4)')
+
+    !Create fake qmin -> will be removed when qmin exists
+    qmin = 0._kind_phys
+
+    !Allocate and initialize arrays whose dimensions depend on num_constituents:
+    num_constituents = number_of_constituents
```
cacraigucar, line 94: I'm not sure if this is a "hard and fast rule" or not, but typically when we set a module variable to an input variable, we use the same name and the input variable has the name appended with "_in".  In this case, it would be `num_constituents = num_constituents_in` or `number_of_constituents = number_of_constituents_in` depending on which name you prefer

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
+module qneg
+
```
cacraigucar, line 2: It appears that this module may just be for CCPP use (and not being taken back into ESCOMP/CAM).  If that is the case, I suggest that a comment be added at the top of this module indicating its limited scope.

cacraigucar (changes requested):
First review.  I wanted to get this off so that Courtney can review my comments and respond.  I may find additional items tomorrow as I'm still comparing the changes.

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
...
+  use shr_kind_mod,        only: CS => SHR_KIND_CS
+  use ccpp_kinds,          only: kind_phys
+  use perf_mod,            only: t_startf, t_stopf
+  use cam_logfile,         only: iulog
+  use shr_sys_mod,         only: shr_sys_flush
+!  use cam_history_support, only: max_fieldname_len
+!  use ppgrid,              only: pcols
```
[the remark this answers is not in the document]
peverwhee: I'll make a pass through and remove some commented out code, but I assume the history stuff will eventually be added back in in some form.

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
...
+
+  implicit none
+  private
+  save
+
+  ! Public interface.
+
```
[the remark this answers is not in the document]
peverwhee: Not 100% sure, but I think not since qneg4 will become a new module named qneg_surface.

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
...
+    !------------------------------Arguments--------------------------------
+    !
+    ! Input arguments
+    !
+    character(len=*), intent(in) :: subnam ! name of calling routine
+
+    integer, intent(in) :: loop_begin
+    integer, intent(in) :: loop_end
+    integer, intent(in) :: lver         ! number of vertical levels in column
+    integer, intent(in) :: num_constituents ! number of constituents
```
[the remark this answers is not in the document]
peverwhee: Steve and I had talked about how, in physics_types,  it's called in a loop over all constituents (with the exception of the number concentrations) so we opted to just loop over all of them within qneg. This issue in CAM - https://github.com/NCAR/CAMDEN/issues/137 - was made to include constituent_properties in qneg and thus handle the number concentrations that way.

utilities/qneg.F90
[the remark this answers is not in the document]
peverwhee: removed one or two things, but left the history stuff commented - let me know if that's something we want to remove altogether and just build from scratch later.

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
...
+  real(kind_phys), parameter :: worst_reset = 1.e35_kind_phys
+
+  ! Diagnostic field names
+!  integer              :: num_diag_fields = 0
+!  character(len=max_fieldname_len) :: diag_names(num_diag_fields)
+!  logical             :: cnst_out_calc = .false.
+!  logical             :: cnst_outfld(num_diag_fields) = .false.
+
+  ! Summary buffers
+  integer, parameter    :: num_bins = 24
```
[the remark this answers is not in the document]
peverwhee: added a comment. Not sure about the first question - @nusbaume ?

utilities/qneg.F90
peverwhee: changed!

utilities/qneg.F90
peverwhee: added a comment.

utilities/qneg.F90
```diff
@@ -0,0 +1,388 @@
...
+!       diag_names(num_constituents+index) = trim(cnst_name(index))//'_qneg_col'
+!       call addfld(diag_names(num_constituents+index), horiz_only, 'I', 'kg/kg',         &
+!            trim(cnst_longname(index))//' QNEG error (column)')
+!    end do
+!    diag_names((2*num_constituents) + 1) = 'qflux_exceeded'
+!    call addfld(diag_names((2*num_constituents) + 1), horiz_only, 'I', 'kg/m^2/s',     &
+!         'qflux excess (QNEG4)')
+
+    !Create fake qmin -> will be removed when qmin exists
+    qmin = 0._kind_phys
```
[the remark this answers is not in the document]
peverwhee: Agreed that, for sure, it will eventually come from the host model. I think the decision was made to put the faked version in here since it doesn't exist anywhere right now and some design decisions have to be made about where it will come from.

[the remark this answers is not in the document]
peverwhee: Added an issue - https://github.com/NCAR/CAMDEN/issues/158 - to CAMDEN

utilities/qneg.F90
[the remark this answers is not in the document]
nusbaume: Assuming the number of schemes that call qneg is equal to the number of qneg calls in a particular SDF then determining the `num_bins` value ahead of time and passing it in should totally be doable.   That being said, although we could get CAMDEN to do it, I think it would be a lot cleaner and more portable if the CCPP framework itself managed it (as @cacraigucar mentioned).  Maybe we should make another issue in this repo to remind us to ask Steve when he gets back?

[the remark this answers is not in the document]
peverwhee: Issue - https://github.com/NCAR/atmospheric_physics/issues/41

utilities/qneg.F90
cacraigucar: Just an FYI - line 79 (currently commented out) uses num_constituents.  So if this logic is ever uncommented, then it will need to be moved below this assignment.

cacraigucar (approved):

utilities/qneg.F90
peverwhee: good call - moved the assignment upwards.

PR #42, opened 2021-09-30
Implement physics updater schemes

peverwhee:
First pass of physics updater schemes
Changed ccpp_error_flag to ccpp_error_code
Updated standard names

Addresses #36

Notes:
- Tested w/ Kessler & Held-Suarez
- It will require a new CCPP Framework tag and updates to CAMDEN standard names

Issue #36, opened 2021-09-10
Need tendency update utility routines


gold2718:
To alleviate the need for specialized state-update routines for each converted scheme, create a set of  `apply_tendency_of_<x>` routines which apply a specific tendency to its corresponding state variable(s).

gold2718:
For the first round, I will assume that `cpair` is a scalar. We need to generalize this once we sort out how we are going to implement thermodynamics in CAMDEN (and CAM). Same goes for other quantities which are approximately constant in the troposphere (e.g., `rair`).

PR #37, opened 2021-09-10
Implement physics updater schemes

gold2718:
First pass of physics updater schemes
Changed ccpp_error_flag to ccpp_error_code
Updated standard names

Addresses #36

Notes:
- This code is not tested yet (which is why it is draft status)
- I modified suite_held_suarez_1994.xml but did not remove held_suarez_update.F90 (that could be part of this PR)
- It will require a new CCPP Framework tag and updates to CAMDEN standard names

PR #45, opened 2021-10-11
Implement physics updater schemes

peverwhee:
First pass of physics updater schemes
Changed ccpp_error_flag to ccpp_error_code
Updated standard names

Addresses #36

Notes:

Implemented & Tested w/ Held-Suarez
It will require a new CCPP Framework tag and updates to CAMDEN standard names

*Remake of #42 since I screwed up and didn't merge up correctly by making a new branch in my own fork.


utilities/physics_tendency_updaters.meta
```diff
@@ -0,0 +1,222 @@
...
+  units = count
+  type = integer
+  dimensions = ()
+  intent = in
+[ heating_rate ]
+  standard_name = heating_rate
+  units = J kg-1 s-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = in
```
cacraigucar, line 127: Based on what @goldy said in #37, I believe there should be a `state_variable=true` here.

cacraigucar: @nusbaume  - We need a scientist to weigh in on this.  Since my understanding of `heating rate` is that it is replacing state%s, I assumed that it was the state variable.  Is this assumption incorrect?   I really haven't gotten a full grasp of what exactly heating rate is yet.

andrewgettelman: @gold2718 had an opinion about this. We have dry static energy (state%s) already in state. My proposal is that we do NOT change it right now, but have parameterizations that want to update it. That I think is how the model works now? I can dive into it if necessary. Only used in:

eddy_diff_cam.F90
physics_types.F90
vertical_diffusion.F90

(also convect_shallow, unicon and subcol_SILHS)

nusbaume: I believe @peverwhee is correct in that this particular variable does NOT need `state_variable=true` here.  The reason is because this quantity isn't actually `state%s`, but just a thermal energy tendency (hence why I requested the name change to `heating_rate` in the first place).  

On top of that I believe this quantity is overwritten by every additional scheme in an SDF that calculates it (similar to something like `ptend_loc` in various CAM physics schemes), which means it also doesn't match my understanding of a state variable from a software design standpoint either.

Of course maybe @gold2718 has a different definition of what a state variable is?

cacraigucar: I am perfectly happy leaving this as it is, no `state_variable=true` for now.  All of the other `apply` routines had their fields as state variables.  With Jesse's explanation of `heating_rate` being a synonym for us non-science types of `tendency of _temperature`, I agree it is not a state variable.

PR #50, opened 2021-11-15
Modify suite name in cam6_silhs SDF to avoid naming conflict.

### LINKED ISSUE #49: Need to remove CAM6 suite name conflict
A new namelist implementation in CAMDEN involves searching for names of suites as specified in the SDFs themselves, as opposed to just using the SDF file names.  This means that two SDFs in this repo cannot have the exact same suite name as specified by the `suite` XML tag. 

Currently this naming conflict only applies to `suite_cam6.xml` and `suite_cam6_silhs.xml`, which both have the `cam6` suite name.  Given that none of these SDFs are actually usable at the moment, the easiest solution would simply be to change the suite name in `suite_cam6_silhs.xml` from `cam6` to `cam6_silhs`.

nusbaume:
Rename physics suite in CAM6-SILHS SDF to avoid naming conflict in CAMDEN.

Fixes #49

### DIFF
--- a/suite_cam6_silhs.xml
+++ b/suite_cam6_silhs.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<suite name="cam6" version="1.0">
+<suite name="cam6_silhs" version="1.0">
   <group name="physics_bc">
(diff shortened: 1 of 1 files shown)


cacraigucar (approved):

gold2718 (approved):
Whew! I'm glad this is the only work this suite needed :)

peverwhee (approved):

PR #55, opened 2022-10-13
Remove erroneous STP label from dry air density.

### LINKED ISSUE #54: dry air density is not at STP
The dry air density variables used in the `state_converters` and `kessler` physics schemes have standard names that imply they are at Standard temperature and Pressure (STP).  However, that density quantity is a function of the atmospheric temperature and pressure, which means it is not in fact at STP.  Thus this label should be removed in order to avoid a name collision with density variables that are in fact at STP.

nusbaume:
See title.

Fixes #54

### DIFF
--- a/kessler/kessler.meta
+++ b/kessler/kessler.meta
@@ -168,7 +168,7 @@
   type = integer
   intent = in
 [ rho ]
-  standard_name = density_of_dry_air
+  standard_name = dry_air_density
   long_name = dry air density
   units = kg m-3
   dimensions = (horizontal_loop_extent, vertical_layer_dimension)
--- a/kessler/kessler.meta
+++ b/kessler/kessler.meta
@@ -168,7 +168,7 @@
   type = integer
   intent = in
 [ rho ]
-  standard_name = density_of_dry_air_at_stp
+  standard_name = density_of_dry_air
   long_name = dry air density
   units = kg m-3
   dimensions = (horizontal_loop_extent, vertical_layer_dimension)
--- a/utilities/state_converters.meta
+++ b/utilities/state_converters.meta
@@ -187,7 +187,7 @@
   dimensions = (horizontal_loop_extent, vertical_layer_dimension)
   intent = in
 [ rho ]
-   standard_name = density_of_dry_air
+   standard_name = dry_air_density
    long_name = dry air density
    units = kg m-3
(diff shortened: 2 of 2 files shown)


peverwhee (approved):

cacraigucar:
An email from @gold2718 :

While I updated the dimension names in the wiki, I am not sure all the variables are there and/or correct. For the most part, I prefer the names used in the Kessler suite. For consistency I suggest the following changes to Kessler:

dry_air_density                                 ==> density_of_dry_air_at_STP

I'm not sure why he chose this and if he has an opinion on our proposed fix

cacraigucar (approved):

nusbaume:
@gold2718 Since you were the one who made the original naming decision I decided to add you as a reviewer, just to be completely safe (even though I seem to be failing on the "reduce your workload" front).

kessler/kessler.meta
```diff
@@ -168,7 +168,7 @@
   type = integer
   intent = in
 [ rho ]
-  standard_name = density_of_dry_air_at_stp
+  standard_name = density_of_dry_air
```
gold2718, line 171: Following the Standard Names dictionary (https://github.com/ESCOMP/CCPPStandardNames/blob/main/Metadata-standard-names.md), this should probably be: `dry_air_density`

utilities/state_converters.meta
```diff
@@ -187,7 +187,7 @@
   dimensions = (horizontal_loop_extent, vertical_layer_dimension)
   intent = in
 [ rho ]
-   standard_name = density_of_dry_air_at_stp
+   standard_name = density_of_dry_air
```
gold2718, line 190: Following the Standard Names dictionary (https://github.com/ESCOMP/CCPPStandardNames/blob/main/Metadata-standard-names.md), this should probably be: `dry_air_density`

gold2718 (changes requested):
(no, I have no idea what I was thinking two years ago when I sent that email).

kessler/kessler.meta
nusbaume: Good call!  Name changed.

utilities/state_converters.meta
nusbaume: Name changed.

gold2718 (approved):
:+1:

PR #58, opened 2023-01-12
fix constituent stdnames in state_converters.meta

peverwhee:
There is a mismatch in standard names for the following:

- cloud_liquid_water_mixing_ratio_wrt_moist_air vs cloud_liquid_water_mixing_ratio_of_moist_air
- rain_water_mixing_ratio vs rain_water_mixing_ratio_wrt_moist_air

The only place that the first names were used was in state_converters.meta. Switching those to the second names listed for consistency. The corrected names match those listed in https://github.com/NCAR/CAMDEN/wiki/Metadata-standard-names

### DIFF
--- a/utilities/state_converters.meta
+++ b/utilities/state_converters.meta
@@ -379,7 +379,7 @@
   dimensions = (horizontal_loop_extent, vertical_layer_dimension)
   intent = in
 [ qc ]
-  standard_name = cloud_liquid_water_mixing_ratio_wrt_moist_air
+  standard_name = cloud_liquid_water_mixing_ratio_of_moist_air
   long_name = Mass mixing ratio of cloud liquid water / moist_air
   advected = True
   type = real | kind = kind_phys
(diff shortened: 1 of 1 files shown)


cacraigucar (approved):

cacraigucar (changes requested):
Oops - so much for doing a review on my phone when I couldn't sleep.  I was "seeing" the changes reversed.

All variables should be using the "wrt"

nusbaume:
Agree with @cacraigucar , as the `wrt` versions follow the CCPP standard name rule number 5 here:

https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst#ccpp-standard-name-rules

However, it does look like we need to update the CAMDEN standard names list (whoops!), so I'll go ahead and do that now.

nusbaume:
Hi All, just letting you know that I have updated the standard names page on CAMDEN to better take into account the CCPP standard name rules for constituents (at least for the ones associated with water).  If you find anything wrong, or have any questions or concerns, please let me know.  Thanks!

peverwhee:
Thanks all! Closing this - will fix on the CAMDEN side

PR #67, opened 2023-09-18
 Update standard and function names plus major code clean-up

### LINKED ISSUE #60: Required atmospheric physics updates
After going through the CAM names with scientists, the following standard names need to be updated:

1. `reference_pressure_at_surface`             -> `reference_pressure`
2. `rain_water_mixing_ratio_wrt_moist_air` -> `rain_mixing_ratio_wrt_moist_air`
3. `rain_water_mixing_ratio_wrt_dry_air`    ->  `rain_mixing_ratio_wrt_dry_air`
4. `ln_of_air_pressure_at_interface`             -> `ln_air_pressure_at_interface` 
5. `pressure_thickness`                                -> `air_pressure_thickness`
6. `pressure_thickness_of_dry_air`               -> `air_pressure_thickness_of_dry_air`
7. `geopotential_at_surface`                       -> `surface_geopotential`
8. `reciprocal_of_pressure_thickness`          -> `reciprocal_of_air_pressure_thickness`
9. `x_wind`                                                   -> `eastward_wind`
10. `y_wind`                                                 -> `northward_wind`
11. `geopotential_height`                            -> `geopotential_height_wrt_surface`
12. `tendency_of_x_wind`                             -> `tendency_of_eastward_wind`
13. `tendency_of_x_wind_due_to_model_physics` -> `tendency_of_eastward_wind_due_to_model_physics`
14. `tendency_of_y_wind`                             -> `tendency_of_northward_wind`
15. `tendency_of_y_wind_due_to_model_physics` -> `tendency_of_northward_wind_due_to_model_physics`

Also a `ChangeLog` file needs to be added that records the modifications made for each new tag, starting with the one created to fix this issue.

### LINKED ISSUE #61: "calc_exner_run" routine should be given reference pressure argument
The `calc_exner_run` subroutine in `state_converters.F90` currently uses a hard-coded value for the Exner reference pressure.  However, this can cause mis-matches with the host model if a different reference pressure is being used elsewhere.  Thus to avoid this the reference pressure should be an argument that is passed in to `cacl_exner_run`, which ideally will come from the host model itself, or at least be consistent with the rest of the schemes in the physics suite.

### LINKED ISSUE #62: Need to update geopotential_t to use "generalized" virtual temperature
Currently in CAM the calculation of geopotential height uses "generalized" virtual temperature, which accounts for all of the possible thermodynamically active constituents that the model simulates.  However, in the CCPP-ized version of `geopotential_t` only water vapor is used, resulting in a mismatch between CAM and CAMDEN (or any other CCPP-enabled model which uses the CCPP-ized `geopotential_t` scheme).

To remedy this problem, the CCPP constituents array needs to be passed in along with information on which constituents are "thermodynamically active".  This information should probably be contained within the `ccpp_constituent_properties_array` structure provided by capgen via https://github.com/NCAR/ccpp-framework/pull/468.

### LINKED ISSUE #64: held_suarez_1994 SDF needs the `geopotential_t` scheme.
Without it not only does it not match what is happening in CAM, but with the latest version of the CCPP-framework the SE dycore is no longer able to properly link with the `geopotential_t_run` subroutine needed to calculate derived quantities in the dynamics-physics coupling layer.

nusbaume:
The main purpose of this PR is to update the CCPP standard names to match the internal AMP agreed-upon names, as well as rename some of the physics routines to make them less ambigious and more accurate.  Along with these changes, significant code cleanup was performed along with some minor bug fixes, with the goal being to produce bit-for-bit results using the same repo version in both CAM and CAM-SIMA.

Fixes #60 
Fixes #61 
Fixes #62 
Fixes #64

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -4,7 +4,7 @@ Tag name: atmos_phys0_01_000
 Originator(s): nusbaume
 Date: 9 Sep 2023
 One-line Summary: Update standard and function names plus major code clean-up
-Github PR URL: https://github.com/NCAR/atmospheric_physics/pull/XXX
+Github PR URL: https://github.com/NCAR/atmospheric_physics/pull/67
 
 Purpose of changes (include the issue number and title text for each relevant GitHub issue):
 
@@ -24,65 +24,49 @@ This PR fixes the following NCAR/atmospheric_physics Github issues:
 
 Code reviewed by: cacraig, peverwhee, mwaxmonsky
 
-List all files eliminated:
+List all existing files that have been added (A), modified (M), or deleted (D),
+and describe the changes:
 
-utilities/geopotential_t.F90 - Replaced with utilities/geopotential_temp.F90
-utilities/geopotential_t.meta - Replaced with utilities/geopotential_temp.meta
-
-List all files added and what they do:
-
-utilities/geopotential_temp.F90 - New geopotential calculation routine which uses
-                                  generalized virtual temperature (i.e. all
-                                  thermodynamically active species).
-utilities/geopotential_temp.meta - Metdata file for new "geopotential_temp.F90" CCPP scheme.
-
-List all existing files that have been modified, and describe the changes:
-
-held_suarez/held_suarez_1994.F90         - Code cleanup
-held_suarez/held_suarez_1994.meta        - Updated variable standard names and removed unused variables
-kessler/kessler.F90                      - Code cleanup and bug fixes
-kessler/kessler.meta                     - Updated variable standard names and removed unused variables
-kessler/kessler_update.F90               - Code cleanup
-kessler/kessler_update.meta              - Updated variable standard names and did some metadata cleanup
-suite_held_suarez_1994.xml               - Updated scheme names, and modified scheme list to better match CAM.
-suite_kessler.xml                        - Updated scheme names.
-utilities/physics_tendency_updaters.F90  - Updated variable names and dimensioning
-utilities/physics_tendency_updaters.meta - Updated variable standard names
-utilities/qneg.F90                       - Code cleanup
-utilities/qneg.meta                      - Updated variable standard names and removed unused variables
-utilities/state_converters.F90           - Code cleanup and updated variable/scheme names.
-utilities/state_converters.meta          - Updated variable standard names and scheme names,
-                                           and removed unused variables.
-utilities/static_energy.F90              - Code cleanup and updated variable dimensioning
-utilities/static_energy.meta             - Updated variable standard names
-
-List all standard names currently not in the CCPPStandardNames dictionary:
-
-    - composition_dependent_specific_heat_of_dry_air_at_constant_pressure
-    - reference_pressure_in_atmosphere_layer_normalized_by_reference_pressure
-    - heating_rate
-    - tendency_of_northward_wind
-    - composition_dependent_ratio_of_dry_air_gas_constant_to_specific_heat_of_dry_air_at_constant_pressure
-    - tendency_of_eastward_wind
-    - scheme_name
-    - rain_mixing_ratio_wrt_dry_air
-    - density_of_dry_air
-    - density_of_fresh_liquid_water_at_0c
-    - composition_dependent_gas_constant_of_dry_air
-    - tendency_of_northward_wind_due_to_model_physics
-    - tendency_of_eastward_wind_due_to_model_physics
-    - number_of_ccpp_constituents
-    - flag_for_mpi_root
-    - ccpp_constituent_minimum_values
-    - log_output_unit
-    - ccpp_constituents
-    - ccpp_constituent_properties
-    - print_qneg_warn
-    - lagrangian_vertical_coordinate
-    - ratio_of_water_vapor_gas_constant_to_composition_dependent_dry_air_gas_constant_minus_one
-    - geopotential_height_wrt_surface_at_interface
-    - rain_mixing_ratio_wrt_moist_air_and_condensed_water
-    - cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water
+A       doc/ChangeLog
+  - So future developers can know the development history
+M       held_suarez/held_suarez_1994.F90
+  - Code cleanup
+M       held_suarez/held_suarez_1994.meta
+  - Updated variable standard names and removed unused variables
+M       kessler/kessler.F90
+  - Code cleanup and bug fixes
+M       kessler/kessler.meta
+  - Updated variable standard names and did some metadata cleanup
+M       kessler/kessler_update.F90
+  - Code cleanup
+M       kessler/kessler_update.meta
+  - Updated variable standard names and did some metadata cleanup
+M       suite_held_suarez_1994.xml
+  - Updated scheme names, and modified scheme list to better match CAM
+M       suite_kessler.xml
+  - Updated scheme names
+D       utilities/geopotential_t.F90
+  - Replaced with utilities/geopotential_temp.F90
+A       utilities/geopotential_temp.F90
+  - New geopotential calculation routine which uses generalized virtual temperature,
+    i.e. all thermodynamically active species
+M       utilities/physics_tendency_updaters.F90
+  - Updated variable names and dimensioning
+M       utilities/physics_tendency_updaters.meta
+  - Updated variable standard names
+M       utilities/qneg.F90
+  - Code cleanup
+M       utilities/qneg.meta
+  - Updated variable standard names and removed unused variables
+M       utilities/state_converters.F90
+  - Code cleanup and updated variable/scheme names
+M       utilities/state_converters.meta
+  - Updated variable standard names and scheme names,
+    and removed unused variables.
+M       utilities/static_energy.F90
+  - Code cleanup and updated variable dimensioning
+M       utilities/static_energy.meta
+  - Updated variable standard names
 
 List and Describe any test failures: No known test failures.
 
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,0 +1,93 @@
+===============================================================
+
+Tag name: atmos_phys0_01_000
+Originator(s): nusbaume
+Date: 9 Sep 2023
+One-line Summary: Update standard and function names plus major code clean-up
+Github PR URL: https://github.com/NCAR/atmospheric_physics/pull/XXX
+
+Purpose of changes (include the issue number and title text for each relevant GitHub issue):
+
+The main purpose of this PR is to update the CCPP standard names to match the
+internal AMP agreed-upon names, as well as rename some of the physics routines
+to make them less ambigious and more accurate.  Along with these changes,
+significant code cleanup was performed along with some minor bug fixes,
+with the goal being to produce bit-for-bit results using the same repo
+version in both CAM and CAM-SIMA.
+
+This PR fixes the following NCAR/atmospheric_physics Github issues:
+
+#60 -> Required atmospheric physics updates
+#61 -> "calc_exner_run" routine should be given reference pressure argument
+#62 -> Need to update geopotential_t to use "generalized" virtual temperature
+#64 -> held_suarez_1994 SDF needs the geopotential_t scheme.
+
+Code reviewed by: cacraig, peverwhee, mwaxmonsky
+
+List all files eliminated:
+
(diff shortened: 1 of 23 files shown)


peverwhee (approved):
Looks good (and like a lot of work!) to me!

Thank you!

nusbaume:
@PeterHjortLauritzen I mostly just requested your review here to see if the updated standard names and physics routine names sound reasonable to you.  However if you end up not having the time to review then just let me know and I'll take you off the review list.  Thanks!

utilities/state_converters.F90
```diff
@@ -187,20 +124,20 @@ subroutine wet_to_dry_water_vapor_run(ncol, nz, pdel, pdeldry, qv, qv_dry,  &
 
      integer,          intent(in)  :: ncol
```
PeterHjortLauritzen, line 125: The conversion from wet to dry and dry to wet mixing ratios is general. I am curious why there are separate calls for water vapor, rain, etc. to convert from wet to dry. Could we use a generic subroutine instead?

PeterHjortLauritzen (commented):
Looks great. Thanks Jesse ... just have one question (see comment for specific code section)

doc/ChangeLog
```diff
@@ -0,0 +1,93 @@
...
+suite_kessler.xml                        - Updated scheme names.
+utilities/physics_tendency_updaters.F90  - Updated variable names and dimensioning
+utilities/physics_tendency_updaters.meta - Updated variable standard names
+utilities/qneg.F90                       - Code cleanup
+utilities/qneg.meta                      - Updated variable standard names and removed unused variables
+utilities/state_converters.F90           - Code cleanup and updated variable/scheme names.
+utilities/state_converters.meta          - Updated variable standard names and scheme names,
+                                           and removed unused variables.
+utilities/static_energy.F90              - Code cleanup and updated variable dimensioning
+utilities/static_energy.meta             - Updated variable standard names
```
cacraigucar, line 57: The ESCOMP/CAM formatting for the ChangeLog file changes might be handier.  We copy/paste the git status listings and then put the description under the line.

kessler/kessler.meta
```diff
@@ -78,85 +66,98 @@
...
+  long_name = air potential temperature
   units = K
   dimensions = (horizontal_loop_extent, vertical_layer_dimension)
   type = real | kind = kind_phys
   intent = inout
 [ qv ]
   standard_name = water_vapor_mixing_ratio_wrt_dry_air
   long_name = mass mixing ratio of water vapor / dry air
-  advected = true
+  advected = True
```
cacraigucar, line 123: Just curious - does case matter for this variable?

cacraigucar (changes requested):
review of ChangeLog and standard names.  Stiil need to review code

doc/ChangeLog
```diff
@@ -0,0 +1,93 @@
...
+    - rain_mixing_ratio_wrt_dry_air
+    - density_of_dry_air
+    - density_of_fresh_liquid_water_at_0c
+    - composition_dependent_gas_constant_of_dry_air
+    - tendency_of_northward_wind_due_to_model_physics
+    - tendency_of_eastward_wind_due_to_model_physics
+    - number_of_ccpp_constituents
+    - flag_for_mpi_root
+    - ccpp_constituent_minimum_values
+    - log_output_unit
```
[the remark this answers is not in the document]
gold2718: Here is an idea (not to be implemented here). Right now, we have `ccpp_error_message` and `ccpp_error_code`. The framework only checks `ccpp_error_code` for non-zero status to decide if it needs to return early. It could also collect any content in `ccpp_error_message` when `ccpp_error_code == 0` and return all such output if the final `ccpp_error_code == 0`. Then the host model could log the output in such cases. If this is not obviously a terrible idea, I can help flesh out the concept in the CCPP Framework repo.

utilities/state_converters.F90
nusbaume: We currently have these separate calls for legacy reasons, i.e. we had converted Kessler before we had any general way to deal with constituents in the framework.  

However, I agree that we should have generic wet <-> dry conversion routines, and now that we have a constituents object that should be doable.  I can make an issue about it so that we remember to tackle it at some point in the future.

nusbaume: I just created the issue, which can be found here: #68

doc/ChangeLog
```diff
@@ -0,0 +1,93 @@
+===============================================================
+
```
cacraigucar, line 2: As I was working on a CAM ChangeLog entry, I realized that we need a ChangeLog_template as well, to provide the blank template that we will fill in for future tags.

PeterHjortLauritzen (approved):

utilities/geopotential_temp.F90
```diff
@@ -0,0 +1,207 @@
...
+      active_flag = .false.
+      do cidx = 1,ncnst
+         !Check if constituent is thermodynamically active:
+         call cprops(cidx)%is_thermo_active(active_flag)
+         if (active_flag) then
+            do klyr = layer_surf, layer_toa, lyr_step
+               do icol = 1, ncol
+                  sum_dry_mixing_ratio(icol, klyr) =                          &
+                  sum_dry_mixing_ratio(icol, klyr) +                          &
+                  carr(icol, klyr, cidx) * qfac(icol,klyr)
```
cacraigucar, line 162: I suggest you indent the continuation lines a few spaces so it is more readily apparent that the assignment carries over for a few lines.  You don't need to indent all the way to the = sign, but a few spaces would better offset this assignment.

cacraigucar (changes requested):
Minor comment for code.  This completes my review of the Fortran

doc/ChangeLog
cacraigucar: I will add, the command I use to get the changes for the ChangeLog is `git diff --name-status SHA` where SHA is the sha of the last CAM tag commit.

doc/ChangeLog
```diff
@@ -0,0 +1,93 @@
...
+utilities/physics_tendency_updaters.meta - Updated variable standard names
+utilities/qneg.F90                       - Code cleanup
+utilities/qneg.meta                      - Updated variable standard names and removed unused variables
+utilities/state_converters.F90           - Code cleanup and updated variable/scheme names.
+utilities/state_converters.meta          - Updated variable standard names and scheme names,
+                                           and removed unused variables.
+utilities/static_energy.F90              - Code cleanup and updated variable dimensioning
+utilities/static_energy.meta             - Updated variable standard names
+
+List all standard names currently not in the CCPPStandardNames dictionary:
```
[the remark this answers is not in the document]
nusbaume: I have added a new `NamesNotInDictionary.txt` file under the `doc` folder that contains the direct output from the stdname checking script.  I have also modified the script to output a timestamp so that in the future we can easily check whether or not the file is up-to-date.

held_suarez/held_suarez_1994.meta
```diff
@@ -65,57 +41,75 @@
...
+[ cappa ]
+  standard_name = composition_dependent_ratio_of_dry_air_gas_constant_to_specific_heat_of_dry_air_at_constant_pressure
+  units = 1
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  type = real | kind = kind_phys
+  intent = in
+[ cpair ]
+  standard_name = composition_dependent_specific_heat_of_dry_air_at_constant_pressure
+  units = J kg-1 K-1
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
```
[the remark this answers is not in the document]
nusbaume: Great catch!  They should be `horizontal_loop_extent`.  I've gone through and replaced all of the `horizontal_dimension` uses with `horizontal_loop_extent` in all scheme run phases.

kessler/kessler.meta
```diff
@@ -4,35 +4,23 @@
...
-[ rair_in ]
-  standard_name = gas_constant_of_dry_air
-  units =  J kg-1 K-1
-  dimensions = ()
-  type = real | kind = kind_phys
-  intent = in
 [ rhoqr_in ]
-  standard_name = density_of_liquid_water_at_0c
-  long_name = density of liquid water at 0C
+  standard_name = density_of_fresh_liquid_water_at_0c
```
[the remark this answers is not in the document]
nusbaume: Good find!  Replaced my standard name with the official one.

doc/ChangeLog
```diff
@@ -0,0 +1,93 @@
...
+
+    - composition_dependent_specific_heat_of_dry_air_at_constant_pressure
+    - reference_pressure_in_atmosphere_layer_normalized_by_reference_pressure
+    - heating_rate
+    - tendency_of_northward_wind
+    - composition_dependent_ratio_of_dry_air_gas_constant_to_specific_heat_of_dry_air_at_constant_pressure
+    - tendency_of_eastward_wind
```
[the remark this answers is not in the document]
nusbaume: I don't think that's true.  The extension list is to help with generating standard names, but in the end the standard name still needs to be added to the dictionary.  Right now there are at least a dozen `tendency_of_X` standard names listed in the dictionary for both NCAR and NOAA variables.  Plus if we don't add the names to the dictionary then (automatically) searching for the names themselves could get significantly harder.

doc/ChangeLog
```diff
@@ -0,0 +1,93 @@
...
+    - scheme_name
+    - rain_mixing_ratio_wrt_dry_air
+    - density_of_dry_air
+    - density_of_fresh_liquid_water_at_0c
+    - composition_dependent_gas_constant_of_dry_air
+    - tendency_of_northward_wind_due_to_model_physics
+    - tendency_of_eastward_wind_due_to_model_physics
+    - number_of_ccpp_constituents
+    - flag_for_mpi_root
```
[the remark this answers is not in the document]
nusbaume: I believe all variables need to be added to the dictionary regardless if they are extensions or not.  The extension instructions are simply to help assist with generating new standard names.

doc/ChangeLog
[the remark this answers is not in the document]
nusbaume: We discussed this at the CCPP-framework meeting on Sept. 25, and it sounds like for now our method is fine (NOAA is apparently violating this rule all over the place), and that once the latest round of capgen PRs has been merged folks will start work on implementing a logging capability in the capgen branch.  I believe all discussion on this is supposed to happen here:

https://github.com/NCAR/ccpp-framework/issues/439

@gold2718  I went ahead and quoted your idea in that issue along with one concern I had.  Feel free to reply there if you have any other thoughts/opinions.  Thanks!

kessler/kessler.meta
nusbaume: I don't think so, but in core python one must use `True` or `False`, so I decided it would be good to match that just in case.

utilities/geopotential_temp.F90
nusbaume: Agreed, normally I try to do that myself, but for whatever reason didn't here.  It should be indented now.

doc/ChangeLog
cacraigucar: If you really want to match CAM, there is a separate section for added, deleted and modified code.  I don't have a strong preference on whether we should have them separate or not.

nusbaume: Yeah I might push back some on that request, given that with this setup one can just copy/paste the `git diff` result directly without having to split it up into different sections.  Mostly I don't want to make future developers/SEs jump through an extra hoop just to "match CAM", when git already provides all of the information as-is. 

However if others feel really strongly about this then I am happy to at least have a discussion about it.

cacraigucar (approved):

held_suarez/held_suarez_1994.meta
```diff
@@ -65,57 +41,75 @@
...
 [ clat ]
   standard_name = latitude
   units = radians
   type = real | kind = kind_phys
   dimensions = (horizontal_loop_extent)
   intent = in
+[ cappa ]
+  standard_name = composition_dependent_ratio_of_dry_air_gas_constant_to_specific_heat_of_dry_air_at_constant_pressure
+  units = 1
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
```
cacraigucar, line 59: Oops- maybe I was too hasty on my approval.  When looking at the ESCOMP/CAM PR, I looked back at this PR, and see that this does not have horizontal_loop_dimension for this run method.

cacraigucar (changes requested):

cacraigucar (approved):

mwaxmonsky (approved):
Looks good to me!  Great work @nusbaume!

Issue #66, opened 2023-08-11
Add ccpp'ized ZM


cacraigucar:
Add CCPP layers to ZM schemes

Use https://github.com/climbfuji/ccpp-physics/tree/feature/iap_dom as a guide

See:  https://docs.google.com/document/d/1jHLVbUCUhdEtY8puVV5FRnnj0yWdlrrw7j1jpNBnMa4/edit for info on why it can not be used as a dropin

PR #69, opened 2023-09-29
Cleanup of ZM for conversion

cacraigucar:
This PR saves the separated routines and the CAM "use" statements have been removed.  This is the first step in the CCPP'ization of ZM.  The purpose of this PR is to begin having CAM use ZM from this repo (and hence needs an official tag).   Further refinement is expected.

Partially addresses #66


nusbaume (approved):
Looks good!  Left a bunch of comments, but mostly just so we can remember to deal with some additional issues and concerns later before finishing the full CCPP conversion needed for CAM-SIMA.

zm/zm_convr.F90
```diff
@@ -0,0 +1,3108 @@
...
+! Author:guang jun zhang, m.lazare, n.mcfarlane. CAM Contact: P. Rasch
+!
+! This is contributed code not fully standardized by the CAM core group.
+! All variables have been typed, where most are identified in comments
+! The current procedure will be reimplemented in a subsequent version
+! of the CAM where it will include a more straightforward formulation
+! and will make use of the standard CAM nomenclature
+!
+!-----------------------------------------------------------------------
+   use phys_control, only: cam_physpkg_is
```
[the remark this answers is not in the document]
gold2718: Not yet but there could be? CAMDEN has the `physics_suite` variable (and associated function in the runtime object). This name is checked against the list in `ccpp_physics_suite_list` from the CCPP host cap.
I suggest adding a new standard name (with request to add it to the [dictionary](https://github.com/ESCOMP/CCPPStandardNames)), `ccpp_physics_suite`, which would contain this string. Then add it to a new host table (`phys_init`) in phys_comp.meta. That should make it available to the scheme (if it adds it as an input).
Thoughts?

nusbaume:
@cacraigucar I just realized that I forgot to include in my review a reminder to add an atmospheric_physics `ChangeLog` entry before this is merged.  Thanks!

zm/zm_convr.F90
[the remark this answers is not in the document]
nusbaume: @gold2718 that sounds like a good general solution to me!

However, after thinking about this specific case some more I wonder if a possibly better solution for the ZM scheme is to just add a new namelist variable to turn on/off the non-dilute buoyancy used in CAM3.  

I feel doing it that way will make this particular scheme more portable (i.e. it won't be looking for a specific CAM-SIMA configuration).  We can then always set the default namelist value to be dependent on the physics suite one is using, which is a feature CAM-SIMA has already.

zm/zm_conv_convtran.F90
```diff
@@ -0,0 +1,318 @@
+module zm_conv_convtran_mod
```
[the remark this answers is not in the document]
gold2718: FYI, this rule (module name conforms to the filename) is in the [Fortran Coding Standards](https://github.com/ESCOMP/CAM/wiki/CAM-Coding-Standards#fortran-coding-standards) under **SHOULD**.

nusbaume:
@cacraigucar Just adding a request here to see if you could change the standard name `lagrangian_vertical_coordinate` in `utilities/geopotential_temp.meta` to `do_lagrangian_vertical_coordinate`?  I realized I had added the wrong standard name to this physics scheme in the last PR.  Thanks!

zm/zm_convr.F90
```diff
@@ -0,0 +1,3108 @@
...
+! Purpose:
+! <Say what the routine does>
+!
+! Method:
+! <Describe the algorithm(s) used in the routine.>
+! <Also include any applicable external references.>
+!
+! Author: phil rasch dec 19 1995
+!
+!-----------------------------------------------------------------------
```
mwaxmonsky, line 2978: Not sure if its possible but if we could find someone that can document this section as well, that might be worthwhile.

zm/zm_convr.F90
```diff
@@ -0,0 +1,3108 @@
...
+! This is contributed code not fully standardized by the CCM core group.
+!
+! this code is very much rougher than virtually anything else in the CCM
+! there are debug statements left strewn about and code segments disabled
+! these are to facilitate future development. We expect to release a
+! cleaner code in a future release
+!
+! the documentation has been enhanced to the degree that we are able
+!
+!-----------------------------------------------------------------------
```
mwaxmonsky, line 2052: Similar documentation comment here.

mwaxmonsky (approved):
Just one comment on a use statement in zm_convr.F90 and a few documentation points that are outside the scope of this PR but figured they'd be worth mentioning while we were looking at it but no reason to hold things up due to those.

Great job @cacraigucar, looks good to me!

zm/zm_convr.F90
```diff
@@ -364,79 +361,79 @@ subroutine zm_convr_run(     ncol    ,pcols   ,pver    , &
...
-   real(kind_phys) hmn(pcols,pver)            ! wg moist static energy.
-   real(kind_phys) hsat(pcols,pver)           ! wg saturated moist static energy.
-   real(kind_phys) qlg(pcols,pver)
-   real(kind_phys) dudt(pcols,pver)           ! wg u-wind tendency at gathered points.
-   real(kind_phys) dvdt(pcols,pver)           ! wg v-wind tendency at gathered points.
-!      real(kind_phys) ud(pcols,pver)
-!      real(kind_phys) vd(pcols,pver)
+   real(kind_phys) dqdt(ncol,pver)           ! wg mixing ratio tendency at gathered points.
+   real(kind_phys) dsdt(ncol,pver)           ! wg dry static energy ("temp") tendency at gathered points.
+!      real(kind_phys) alpha(ncol,pver)      ! array of vertical differencing used (=1. for upstream).
```
nusbaume, line 408: Should we go ahead and remove the commented-out variables here and below?

zm/zm_convr.F90
```diff
@@ -1907,7 +1606,8 @@ SUBROUTINE ientropy (rcall,icol,s,p,qt,T,qst,Tfg,cpliq,cpwv,rh2o,errmsg,errflg)
 ! for T and saturated vapor mixing ratio
 !
 
-  use phys_grid, only: get_rlon_p, get_rlat_p
+! CACNOTE - Remove this when pass in lat/lon or pass out lchnk,icol
+!  use phys_grid, only: get_rlon_p, get_rlat_p
```
nusbaume, line 1610: I am not sure if passing out `lchnk,icol` will work because there is no `lchnk` in CAM-SIMA.  Just an FYI to future you.

nusbaume (approved):
The new updates look good to me!  Just a couple of notes (which can be dealt with in the next round if preferred).

zm/zm_convr.F90
cacraigucar: deleted

zm/zm_convr.F90
cacraigucar: Noted this in my comment in the code

mwaxmonsky (commented):
Great work @cacraigucar!  Just some commented out code that I think we can safely delete based on discussions with @nusbaume.

The other thing I was wondering if there's any value to is looking at the formatting?  There's several sections of code that change formatting ad-hoc and was wondering if we could make it consistent with the rest of the SIMA code base before we merge?  It might take a bit of working through and I'd be happy to help and I think it would be valuable to start with as clean of a slate as possible.

cacraigucar:
@swrneale - Are you okay with the removal of all the commented out lines that @mwaxmonsky has indicated?

mwaxmonsky:
@cacraigucar Thanks for taking care of the commented out code!

I was just wondering if there was any discussion on the formatting side of things?  Don't think that should hold up this PR and we can open a new one if it's worth pursuing but if it's reasonable, I'd be happy to help with that as well.

cacraigucar:
> @cacraigucar Thanks for taking care of the commented out code!
> 
> I was just wondering if there was any discussion on the formatting side of things? Don't think that should hold up this PR and we can open a new one if it's worth pursuing but if it's reasonable, I'd be happy to help with that as well.

No there was no discussion on formatting.  I left it in the weird state that it is.  (I assume you are talking about the spaces in the calling lists, but I suppose there may be other locations as well).  At this point, if you'd like to make the formatting edits for the next time around, feel free.   I'd like this code to look as close to the CAM version so when @rneale brings in his changes, it is apparent to see where they go in relation to his personal fork.

cacraigucar:
Also, I should point out that I am deferring removing the zm_org logic as well with this commit.

swrneale:
Also I will work on finally expanding the org infrastructure as per this [pre-historic](https://github.com/ESCOMP/CAM/issues/190) issue over on ESCOMP

Issue #70, opened 2023-10-16
Update MICM interface to run simple MICM configuration


mattldawson:
Update the MICM CCPP wrapper to use new MUSICA functions introduced with https://github.com/NCAR/musica/pull/35

Have the wrapper create an instance for some very simple chemical system and solve at each timestep based on fixed conditions (no use of or updating of the CAM-SIMA state)

PR #71, opened 2023-11-03
Update the MICM CCPP wrapper to use new MUSICA functions

boulderdaze:
Closes [70](https://github.com/ESCOMP/atmospheric_physics/issues/70)

### DIFF
--- a/musica/micm/micm.F90
+++ b/musica/micm/micm.F90
@@ -47,34 +47,44 @@ module micm
    !> \section arg_table_micm_run Argument Table
    !! \htmlinclude micm_run.html
    subroutine micm_run(temperature, pressure, time_step, concentrations, num_concentrations, iulog, errcode, errmsg)
-      real(kind_phys), intent(in)                                    :: temperature
-      real(kind_phys), intent(in)                                    :: pressure
-      real(kind_phys), intent(in)                                    :: time_step
-      real(kind_phys), dimension(:,:,:), allocatable, intent(inout)  :: concentrations      
+      real(kind_phys), dimension(:,:), intent(in)        :: temperature
+      real(kind_phys), dimension(:,:), intent(in)        :: pressure
+      real(kind_phys), intent(in)                        :: time_step
+      real(kind_phys), dimension(:,:,:), intent(inout)   :: concentrations      
+      integer, intent(in)                                :: num_concentrations
+      integer, intent(in)                                :: iulog
+      integer, intent(out)                               :: errcode
+      character(len=512), intent(out)                    :: errmsg
 
-      real(c_double)                                                 :: c_temperature
-      real(c_double)                                                 :: c_pressure
-      real(c_double)                                                 :: c_time_step
-      real(c_double), dimension(:,:,:), allocatable                  :: c_concentrations
-   
-      integer, intent(in)                                            :: num_concentrations
-      integer, intent(in)                                            :: iulog
-      integer, intent(out)                                           :: errcode
-      character(len=512), intent(out)                                :: errmsg
-      
-      real(kind_phys) :: var_phys
-      real(c_double)  :: var_doub
-      
+      ! local variables
+      real(c_double), dimension(shape(temperature))      :: c_temperature
+      real(c_double), dimension(shape(pressure))         :: c_pressure
+      real(c_double)                                     :: c_time_step
+      real(c_double), dimension(shape(concentrations))   :: c_concentrations
+      integer                                            :: horizontal_loop_extent
+      integer                                            :: vertical_layer_dimension
+      integer                                            :: i_column
+      integer                                            :: i_layer
+
       c_temperature = real(temperature, c_double)
       c_pressure = real(pressure, c_double)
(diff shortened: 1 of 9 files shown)


boulderdaze:
Would it be possible to add @mattldawson and @K20shores as reviewers?

peverwhee:
@boulderdaze I sent them invites to collaborate on this repo and once they accept we'll be able to add them as reviewers.

K20shores (approved):

musica/micm/micm.F90
```diff
@@ -1,87 +1,109 @@
...
-         import :: c_ptr, c_double, c_int64_t
-         real(c_double), dimension(*) :: state
-         integer(c_int64_t), value :: state_size
-         integer(c_int64_t), value :: time_step
-      end subroutine
-
-   end interface
-
-contains
+   type(micm_t), pointer :: micm_ptr
```
mattldawson, line 14:
```suggestion
   type(micm_t), pointer :: micm_ptr => null()
```
to make sure `associated()` can be used on `micm_ptr`

musica/micm/micm.F90
```diff
@@ -1,87 +1,109 @@
...
+      character(len=512), intent(out)  :: errmsg
 
-      write(log_output_unit,*), "new state", state_cdouble
+      errcode = 0
+      errmsg = ''
       
-      ! Deallocate the temporary array
-      deallocate(state_cdouble)
+      write(iulog,*) "FINAL MICM: INFO: Deallocating MICM object..."
+      deallocate(micm_ptr)
```
mattldawson, line 104:
```suggestion
      if (associated(micm_ptr)) deallocate(micm_ptr)
```

musica/micm/micm.meta
```diff
@@ -5,56 +5,101 @@
...
   units = count
   type = integer
   dimensions = ()
   intent = in
-[ timestep_for_physics ]
-  standard_name = timestep_for_physics
-  units = s
+[ concentrations ]
+  standard_name = ccpp_constituents
+  units = none
```
mattldawson, line 62: should this be `mol m-3`? (If not, we will have to do some unit conversions in the solve function)

mattldawson (changes requested):
looks great! Just a couple minor things (and I'm not sure how the CCPP constituents object works, so I couldn't really comment on that part)

musica/micm/micm.meta
```diff
@@ -5,56 +5,101 @@
...
   units = count
   type = integer
   dimensions = ()
   intent = in
-[ timestep_for_physics ]
-  standard_name = timestep_for_physics
-  units = s
+[ concentrations ]
+  standard_name = ccpp_constituents
+  units = mol m-3
```
nusbaume, line 62: We are currently trying to figure out what the "best" solution is to this issue of constituents having different units.  For now it sounds like you should assume that the units are in mass mixing ratios of  `kg chemical species / kg of air` (so `units = kg kg-1`), and then add a conversion step to go from `kg kg-1` to `mol m-3`, with the understanding that the conversion code will eventually go away and be handled entirely by the framework itself, if it is needed at all.

@peverwhee Do I have that correct?

musica/micm/micm.meta
```diff
@@ -5,56 +5,101 @@
...
   type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent,vertical_layer_dimension,number_of_ccpp_constituents)
+  intent = inout
+[ iulog ]
+  standard_name = log_output_unit
+  units = none
```
nusbaume, line 68: `log_output_unit` should have units of `1`, as `none` is generally saved for characters and strings:
```suggestion
  units = 1
```

musica/micm/micm.meta
```diff
@@ -5,56 +5,101 @@
...
   intent = out
-[ errflg ]
-  standard_name = errflg
+
+[ccpp-arg-table]
+  name  = micm_final
+  type  = scheme
+[ iulog ]
+  standard_name = log_output_unit
+  units = none
```
nusbaume, line 90: Need to change the units here:

```suggestion
  units = 1
```

nusbaume (changes requested):
Thanks for the great work @boulderdaze!  There are some things which need to be changed, but hopefully most of my change requests are easy to fix.  Of course please let me know if you need any questions or concerns.

musica/micm/micm.meta
nusbaume: This concern should be (at least temporarily) fixed by issue #73, so am marking this as resolved.

nusbaume (approved):
Everything looks good to me now, thanks!

musica/micm/micm.meta
nusbaume: @peverwhee Yes!  My mistake.  @boulderdaze can you change the units listed here to `kg kg-1`?  Otherwise the framework might complain in the future about incompatible units (at least until we have a long-term plan for constituents with different units).

peverwhee (approved):

mattldawson (approved):
looks great!

PR #81, opened 2024-02-14
Unit updates for consistency w/ SIMA

peverwhee:
In bringing in new framework updates, the unit conversion abilities are coming along as well! The result is that unit differences actually matter now. Here are some unit fixes for atmospheric_physics.

PR also brings in standard name updates to align with the dictionary


nusbaume (approved):
Looks great, thanks @peverwhee!

utilities/qneg.meta
```diff
@@ -133,13 +133,13 @@
   type  = scheme
 [ mpi_communicator ]
   standard_name = mpi_communicator
-  units = 1
+  units = index
   type = integer
   dimensions = ()
   intent = in
 [ rootprocid ]
```
cacraigucar, line 140: Isn't this mpi_root (based on what is in the CCPP standard dictionary)?

utilities/qneg.meta
```diff
@@ -61,7 +61,7 @@
   intent = in
 [ q ]
   standard_name = ccpp_constituents
-  units = kg kg-1
+  units = none
```
cacraigucar, line 64: I was told not that long ago (I think) that I should be using  kg kg-1 for ccpp_constituents - that's what I'm doing for the zm code.

cacraigucar (changes requested):
Both of my comments apply to a couple of locations each if changes need to be made

utilities/qneg.meta
peverwhee: I think it needs to be none. That's what's in the dictionary and the units are (at least currently) determined by the constituent properties array

utilities/qneg.meta
peverwhee: I think you're looking at the local name - the standard name is indeed mpi_root

cacraigucar: Yup - you are correct

utilities/qneg.meta
cacraigucar: OK - I'll change my ZM metadata files as well.

cacraigucar (approved):

PR #84, opened 2024-03-06
update qneg namelist standard name

peverwhee:
Update print_qneg_warn standard name in qneg namelist.

...I guess I should have run a test before merging that last PR

### DIFF
--- a/utilities/qneg_namelist.xml
+++ b/utilities/qneg_namelist.xml
@@ -77,11 +77,11 @@
                                                                         -->
 <!-- Qneg Parameterization -->
 
-  <entry id="print_qneg_warn">
+  <entry id="control_for_negative_constituent_warning">
     <type>char*8</type>
     <category>diagnostics</category>
     <group>qneg_nl</group>
-    <standard_name>print_qneg_warn</standard_name>
+    <standard_name>control_for_negative_constituent_warning</standard_name>
     <units>1</units>
     <valid_values>summary,timestep,off</valid_values>
     <desc>


utilities/qneg_namelist.xml
```diff
@@ -77,11 +77,11 @@
                                                                         -->
 <!-- Qneg Parameterization -->
 
-  <entry id="print_qneg_warn">
```
nusbaume, line 80: Should we change `print_qneg_warn` here?  I thought this is what a user would manually add to the `user_nl_cam` file, and clearly `pring_qneg_warn` is shorter than `control_for_negative_constituent_warning`, although it is also less descriptive.  Any thoughts?  Also pinging @cacraigucar to get her thoughts here too.

nusbaume (commented):
Looks good, but I did have one question.

utilities/qneg_namelist.xml
peverwhee: I'm good either way. It's definitely shorter but, on the other hand, using the standard name might make things less complicated (one name for the variable).

cacraigucar: I would vote that we keep the names that users use as namelists not overly verbose (i.e. most CCPP standard names).  Also to me, `print_qneg_warn` is much more intuitive with what I know about the code than `control_for_negative_constituent_warning`

nusbaume (approved):
Everything looks great to me now, thanks!

utilities/qneg_namelist.xml
```diff
@@ -81,7 +81,7 @@
     <type>char*8</type>
     <category>diagnostics</category>
     <group>qneg_nl</group>
-    <standard_name>print_qneg_warn</standard_name>
+    <standard_name>control_for_negative_constituent_warning</standard_name>
```
cacraigucar, line 84: Totally pedantic question - how do we find `control_for_negative_constituent_warning`?  In the spreadsheet, the standard name is "blank", even though the "Accepted" column is filled in.  Going back to previous versions, I see "print_qneg_warn" as being accepted.

cacraigucar:
OK - I see it in the official repo (but I'm not sure if I would have equated it to qneg).  I'll go ahead and add it into the spreadsheet.

cacraigucar (approved):

peverwhee:
@cacraigucar thanks for adding it to the spreadsheet!

PR #109, opened 2024-08-15
Remove 0.5*timestep logic from call to zm

cacraigucar:
ZM has for a very long time passed in .5*timestep and then inside multiplied by 2 in several places (and also done 0.5/input_timestep in several other locations).  It also hinders CCPP'ization as we do not have a half timestep standard name.  This confusing, unnecessary logic will be removed.

To confirm the results, see this email from @adamrher

Your ZM branch has round-off level differences with cam6_4_018, so it's working as expected (plots attached).
Ignore the top right panel "cam6_4_015" -- that's my clubb+mf branch and something is very wrong with that run.

### DIFF
--- a/zhang_mcfarlane/zm_conv_convtran.F90
+++ b/zhang_mcfarlane/zm_conv_convtran.F90
@@ -22,7 +22,7 @@ subroutine zm_conv_convtran_run(ncol, pver, &
                     doconvtran,q       ,ncnst   ,mu      ,md      , &
                     du      ,eu      ,ed      ,dp      ,dsubcld , &
                     jt      ,mx      ,ideep   ,il1g    ,il2g    , &
-                    nstep   ,fracis  ,dqdt    ,dpdry   ,dt      )
+                    nstep   ,fracis  ,dqdt    ,dpdry)
 !-----------------------------------------------------------------------
 !
 ! Purpose:
@@ -67,9 +67,6 @@ subroutine zm_conv_convtran_run(ncol, pver, &
    integer, intent(in) :: nstep          ! Time step index
 
    real(kind_phys), intent(in) :: dpdry(:,:)    ! Delta pressure between interfaces        (ncol,pver)
-
-   real(kind_phys), intent(in) :: dt                      ! 2 delta t (model time increment)
-
 
 ! input/output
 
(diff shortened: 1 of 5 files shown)


cacraigucar:
[temp_PRECC.pdf](https://github.com/user-attachments/files/16626896/temp_PRECC.pdf)

zhang_mcfarlane/zm_conv_convtran.F90
```diff
@@ -68,9 +68,6 @@ subroutine zm_conv_convtran_run(ncol, pver, &
 
    real(kind_phys), intent(in) :: dpdry(:,:)    ! Delta pressure between interfaces        (ncol,pver)
 
-   real(kind_phys), intent(in) :: dt                      ! 2 delta t (model time increment)
```
adamrher, line 71: I presume dt is just not used in convtran?

adamrher (approved):
Looks good to me!

zhang_mcfarlane/zm_conv_convtran.F90
cacraigucar: That is correct - when I went to see if the dt needed adjusting in this routine, I discovered it was not used, so I removed it.

zhang_mcfarlane/zm_conv_momtran.F90
```diff
@@ -59,7 +59,7 @@ subroutine zm_conv_momtran_run(ncol, pver, pverp, &
    real(kind_phys), intent(in) :: ed(:,:)       ! Mass entraining from downdraft            (ncol,pver)
    real(kind_phys), intent(in) :: dp(:,:)       ! Delta pressure between interfaces         (ncol,pver)
    real(kind_phys), intent(in) :: dsubcld(:)       ! Delta pressure from cloud base to sfc  (ncol)
-   real(kind_phys), intent(in) :: dt                   !  time step in seconds : 2*delta_t
+   real(kind_phys), intent(in) :: dt
```
peverwhee, line 62: can we put the `!  time step in seconds` comment back?

peverwhee (approved):
one tiny thing!

zhang_mcfarlane/zm_conv_momtran.F90
adamrher: That's fine by me @cacraigucar

PR #135, opened 2024-10-16
Implement ccpp_const_get_index in to_be_ccppized to avoid circular dependencies

jimmielin:
This routine can be shared by CAM-SIMA code (companion PR to be filled) and `atmospheric_physics` schemes by passing in the CCPP constituents pointer object.

This loop-based workaround is to avoid a dependency on `cam_ccpp_cap` which would in turn depend on all schemes creating a circular dependency.

### DIFF
--- a/to_be_ccppized/ccpp_const_utils.F90
+++ b/to_be_ccppized/ccpp_const_utils.F90
@@ -1,0 +1,50 @@
+! ccpp_const_utils contains utility functions that use
+! the ccpp constituents property pointer.
+! this code was separated out to remove circular dependencies.
+module ccpp_const_utils
+  implicit none
+  private
+
+  public :: ccpp_const_get_idx
+
+contains
+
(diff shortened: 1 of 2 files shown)


to_be_ccppized/ccpp_const_utils.F90
```diff
@@ -0,0 +1,50 @@
+! ccpp_const_utils contains utility functions that use
+! the ccpp constituents property pointer.
```
nusbaume, line 2: Total nit-pick, but I might change the grammar slightly here:

```suggestion
! the ccpp constituent properties pointer.
```

cacraigucar (approved):
I wanted to try using this in ZM, but right now I'm deep in something else, and don't want to "rock the boat" with this code.  So I'm approving this for now, and if I run into difficulties when I go to use it (I don't expect I will), I can open an issue.

PR #138, opened 2024-10-17
Update temperatures in TUV-x before calculating rate constants

### LINKED ISSUE #95: Update temperatures in TUV-x before calculating rate constants
Update the temperatures profile edges in TUV-x for each column prior to calculating rate constants.

**Acceptance Criteria**
- TUV-x photolysis rate constants are calculated for the actual temperatures for each grid cell at each timestep

**Ideas**
- Identify the CCPP standard name for temperature at layers/interfaces
- Map the temperatures to the TUV-x height grid
- Follow the logic of the TUV-x PR into CAM: https://github.com/ESCOMP/CAM/blob/c12d1e46e0fdc1dccb0a651a6c9fefd6bb80b2ba/src/chemistry/mozart/mo_tuvx.F90#L1391-L1413

boulderdaze:
Originator(s): @boulderdaze 

- Update the temperatures profile edges in TUV-x for each column prior to calculating rate constants.
-  Closes #95

- A       schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
- A       test/musica/tuvx/test_tuvx_temperature.F90

List all existing files that have been modified, and describe the changes: N/A
- M       schemes/musica/musica_ccpp.F90
- M       schemes/musica/musica_ccpp.meta
- M       schemes/musica/tuvx/musica_ccpp_tuvx.F90
- M       test/musica/CMakeLists.txt
- M       test/musica/test_musica_api.F90
- M       test/musica/tuvx/CMakeLists.txt
- M       test/musica/tuvx/configs/ts1_tsmlt.json
- M       test/valgrind.supp

List any test failures: N/A

Is this a science-changing update? New physics package, algorithm change, tuning changes, etc? No


schemes/musica/tuvx/musica_ccpp_tuvx.F90
```diff
@@ -62,31 +64,57 @@ subroutine tuvx_init(vertical_layer_dimension, vertical_interface_dimension, &
       return
     end if
 
-    radiators => radiator_map_t( error )
+    temperature_profile => create_temperature_profile( height_grid, errmsg, errcode )
+    if (errcode /= 0) then
```
jimmielin, line 68: Is there a reason why `has_error_occurred` that is used in the subsequent subroutine calls is not used here?

schemes/musica/tuvx/musica_ccpp_tuvx.F90
```diff
@@ -62,31 +64,57 @@ subroutine tuvx_init(vertical_layer_dimension, vertical_interface_dimension, &
...
+      deallocate( grids )
+      deallocate( height_grid )
+      height_grid => null()
+      deallocate( profiles )
```
jimmielin, line 72: There seems to be a lot of code duplication in the `deallocate` / `=> null()` clean-up calls here whenever an error occurs. Would it be possible to wrap this into a clean-up subroutine that deallocates/nullifies the pointers in this module if allocated?

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
```diff
@@ -0,0 +1,70 @@
...
+
+    profile => profile_t( temperature_label, temperature_unit, &
+                          height_grid, error )
+    if ( has_error_occurred( error, errmsg, errcode ) ) return
+
+  end function create_temperature_profile
+
+  !> Sets TUV-x temperature values from host-model temperatures
+  !!
+  !! See description of `musica_ccpp_tuvx_hegihts_grid.F90` for
```
jimmielin, line 39: hegihts -> heights

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
```diff
@@ -0,0 +1,70 @@
...
+    integer,          intent(out)   :: errcode
+
+    ! local variables
+    type(error_t)   :: error
+    real(kind_phys) :: interfaces(size(host_midpoint_temperatures)+2)
+    integer         :: n_host_midpoint_temperatures
+
+    n_host_midpoint_temperatures = size(host_midpoint_temperatures)
+
+    interfaces(1) = host_surface_temperature
```
jimmielin, line 61: Would it be good to document somewhere (even just in a comment) why `blackbody_temperature_at_surface` is used at the lowest TUV vertical level here instead of the air temperature from the host model elsewhere?

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
```diff
@@ -0,0 +1,70 @@
...
+    interfaces(2:n_host_midpoint_temperatures+1) = host_midpoint_temperatures(n_host_midpoint_temperatures:1:-1)
```
jimmielin, line 62: Similarly I'm not sure if there is a note that the TUV-x vertical grid is surface-as-level-1 vs. TOA-as-level-1 which seems to be assumed here from the host midpoint temperatures. If there is somewhere, please feel free to ignore this comment!

I'm not familiar enough with CCPP to know if it is assumed that `air_temperature` is always provided as TOA = level 1 (and thus the values need to be always flipped in the vertical) -- if it is convention then please also ignore!

jimmielin (changes requested):
Thanks for the PR! This looks good, I just had a few minor questions about the code and some assumptions regarding the vertical grid from the host model.

schemes/musica/tuvx/musica_ccpp_tuvx.F90
boulderdaze: I don't have a definitive answer for that. My impression is that this is CCPP-enabled code that cares about error codes and may not recognize `error_t`, but I could be mistaken. Hi @mattldawson , could you clarify this?

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
boulderdaze: I had a similar question regarding what to use for `surface_temperature` according to CCPP standards. Jesse mentioned that we can use `blackbody_temperature_at_surface` for now, assuming that TUV-x needs the temperature everywhere, and this is something that CGD needs to discuss with NOAA. I added [comments](tuvx_run) about this.

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
boulderdaze: Thank you for catching it! Done

schemes/musica/tuvx/musica_ccpp_tuvx.F90
mattldawson: That's correct - in the MUSICA library, functions return an `error_t` instance that `has_error_occurred()` converts into the `errmsg` and `errcode` that CAM-SIMA uses and returns true if an error has occurred. In this case, the `create_temperature_profile()` is a function in the atmospheric physics repo, so it just returns the `errmsg` and `errcode` directly.

schemes/musica/tuvx/musica_ccpp_tuvx.F90
boulderdaze: Yes, we wanted to address that and I've opened a small PR to specifically for that. I'll add you as a reviewer. Would you be able to review it?

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
mattldawson: Yeah, the conversion between TUV-x and CAM-SIMA vertical grids is pretty complicated. We tried to document it here: https://github.com/ESCOMP/atmospheric_physics/blob/c6ed2caae61dee8ba798011fa3eace583f3db917/schemes/musica/tuvx/musica_ccpp_tuvx_height_grid.F90#L8-L43

We can update this as needed if anything is missing

schemes/musica/tuvx/musica_ccpp_tuvx.F90
jimmielin: Thanks for clarifying!

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
jimmielin: Thanks @mattldawson!

schemes/musica/tuvx/musica_ccpp_tuvx.F90
jimmielin: Done, thank you!

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
nusbaume: Just wanted to note here that the reason I suggested the use of `blackbody_temperature_at_surface` is because that is what we currently have as the standard name for `cam_in%ts`, which is what is being used in CAM for this same quantity.  

There is a question that this is maybe similar to `skin_temperature`, which is what NOAA uses in some places, although I believe technically these are different physical quantities.  Either way, I think for now using this standard name will allow the coupling to CAM-SIMA to work appropriately, and we can then always go back and change the standard name itself if we come to an agreement with NOAA.

nusbaume (approved):
Looks good to me!

schemes/musica/tuvx/musica_ccpp_tuvx_temperature.F90
boulderdaze: Thank you for providing additional information! I've updated the comments to include it.

jimmielin: Thank you!

jimmielin (approved):

Issue #123, opened 2024-09-30
Update radiation grid in TUV-x based on that used in CAM-SIMA


mattldawson:
Create a radiation grid for TUV-x to use during creation of wavelength-dependent profiles (e.g., surface albedo, et flux).

**Acceptance Criteria**
- A `"wavelength"` grid with units of `"nm"` is created during TUV-x initialization, with values based on the wavelength grid used in CAM-SIMA for radiation

**Ideas**
- Either read in the grid from a json file or use wavelength bin mid-point and edge values from CAM-SIMA
- Possible logic from the TUV-x in CAM PR may be useful: https://github.com/mattldawson/CAM/blob/402be6eb614ab0ba9710c619dd9e46cef50984f4/src/chemistry/mozart/mo_tuvx.F90#L1075-L1085

PR #142, opened 2024-10-23
Develop-123-tuvx-radiation-grid

mattldawson:
Updates TUV-x wrapper, tests, and configuration data to allow setting the wavelength grid from the host model. This implementation assumes that there will be a standard-named variable that defines the edges of the wavelength grid to use for photolysis rate constant calculations that is defined by the host model.

Once this gets merged into `development` #123 can be closed.

### DIFF
--- a/schemes/musica/musica_ccpp.meta
+++ b/schemes/musica/musica_ccpp.meta
@@ -19,6 +19,12 @@
   type = integer
   dimensions = ()
   intent = in
+[ photolysis_wavelength_grid_interfaces ]
+  standard_name = photolysis_wavelength_grid_interfaces
+  units = m
+  type = real | kind = kind_phys
+  dimensions = (photolysis_wavelength_grid_interface_dimension)
+  intent = in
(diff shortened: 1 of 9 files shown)


schemes/musica/musica_ccpp.meta
```diff
@@ -19,6 +19,12 @@
   type = integer
   dimensions = ()
   intent = in
+[ photolysis_wavelength_grid_interfaces ]
+  standard_name = photolysis_wavelength_grid_interfaces
+  units = m
```
boulderdaze, line 24: In the configuration (`ts1_tsmlt.json`), the unit is `nm`. Are we going to use `m` and to add conversion?

boulderdaze (approved):
Looks good!

schemes/musica/musica_ccpp.meta
boulderdaze: I saw the conversion. never mind!

PR #143, opened 2024-10-24
Add helper function for deallocating objects associated with TUV-x

boulderdaze:
Add helper function for deallocating objects associated with TUV-x


mattldawson (approved):
looks good to me! I like that it makes the init function much shorter, particularly since we still have more profiles, etc. to add

jimmielin (approved):
Looks good, thank you!

Issue #82, opened 2024-02-24
Need to add unit tests for utility schemes


nusbaume:
Given the implementation of a unit testing framework for the atmospheric_physics repo in PR #78, there is now an opportunity to create some unit tests for the various physics schemes inside the `utilities` directory.  

This would be particularly beneficial because the tests done in CAM-SIMA are usually for an entire physics suite, where the utility schemes are grouped together, as opposed to each individual utility scheme as proposed here.  Plus it would allow for additional test coverage for a PR without necessarily requiring the user to run CAM-SIMA.

PR #144, opened 2024-10-24
Unit test infrastructure setup

mwaxmonsky:
Originator(s): @mwaxmonsky 

Summary (include the keyword ['closes', 'fixes', 'resolves'] and issue number): Sets up the basic infrastructure to start enabling unit testing.  Addresses #82

A       .github/workflows/code-coverage.yaml
A       test/unit-test/CMakeLists.txt
A       test/unit-test/include/ccpp_kinds.F90
A       test/unit-test/tests/CMakeLists.txt
A       test/unit-test/tests/utilities/CMakeLists.txt
A       test/unit-test/tests/utilities/test_state_converters.pf
- Initial round of infrastructure to start enabling unit tests.

List all existing files that have been modified, and describe the changes: N/A

List any test failures: N/A

Is this a science-changing update? New physics package, algorithm change, tuning changes, etc?  No


.github/workflows/code-coverage.yaml
```diff
@@ -0,0 +1,70 @@
+name: code-coverage
+
+on:
+  push:
+    branch: development
```
nusbaume, line 5: Is there any reason to not run this on `main` as well?

test/unit-test/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+                  "project yet and this CMake is for testing purposes only.  "
+                  "Making a change to this project's CMake will not impact the build of "
+                  "a parent project at this time.")
+endif()
+
+option(ATMOSPHERIC_PHYSICS_ENABLE_TESTS "Run pFUnit unit tests" OFF)
+option(ATMOSPHERIC_PHYSICS_ENABLE_CODE_COVERAGE "Run code coverage tool" OFF)
+
+if(ATMOSPHERIC_PHYSICS_ENABLE_CODE_COVERAGE)
+  add_compile_options(-O0 --coverage)
```
nusbaume, line 19: This is probably my CMake ignorance, but why add these complier options here when it looks like we are already setting them above with the `set(CMAKE_Fortran_FLAGS)` command?

nusbaume (approved):
Looks good to me!  Just had a couple of questions.

Also, when you get the chance can you add some general instructions for how to add a new unit test to the official CAM-SIMA documentation?  You can probably put it under the "Atmospheric_physics" section for now (but feel free to add it somewhere else if you think it makes more sense).  Thanks!

.github/workflows/code-coverage.yaml
mwaxmonsky: Good call, added!

test/unit-test/CMakeLists.txt
mwaxmonsky: Nope, left over from testing.  Removed!

test/unit-test/tests/utilities/test_state_converters.pf
```diff
@@ -0,0 +1,27 @@
...
+    temp = 1
+    exner = 1
+    theta = 1
+
+    errmsg = ""
+    errflg = 0
+
+    call temp_to_potential_temp_run(ncol, nz, temp, exner, theta, errmsg, errflg)
+
+    @assertEqual(0, errflg)
```
jimmielin, line 25: I may have missed the discussion on this (perhaps from the Monday meeting?) that perhaps it may be useful to use realistic values in these kinds of tests that have known physical basis and thus the formula itself within the parameterization can also be validated in the test? It does not have to be resolved within the context of this PR so I have approved it, but it would be useful for the future. Thanks!

jimmielin (approved):

test/unit-test/tests/utilities/test_state_converters.pf
mwaxmonsky: Yup, you are correct!  We can add those details in here if needed (looking at @nusbaume for guidance on that one) but this was more to get the infrastructure and an example in place.  But if there's a straightforward example that we can validate with this, I'm more than happy to add it as a part of this PR!

mwaxmonsky: So is this change okay as is and then we can work on a future PR to bring in realistic values?

jimmielin: Sounds good to me!

.github/workflows/code-coverage.yaml
```diff
@@ -0,0 +1,72 @@
+name: code-coverage
+
+on:
+  push:
+    branches:
+      - development
+      - main
+  workflow_dispatch:
+  pull_request:
```
peverwhee, line 9: github wofklow question from a newbie - does this mean we won't see the results of the unit tests when a PR is opened? only when merged?

also: is "code-coverage.yaml" the best name for this workflow? It seems like it's doing more than that. Maybe fortran-unit-tests.yaml?

peverwhee (commented):
couple small questions.

Also, I know you've documented this stuff on the new doc page (thanks!), but can I also request a README in the test/unit-test directory that describes, at minimum, how to run the tests?

.github/workflows/code-coverage.yaml
mwaxmonsky: This will run the workflow when the PR is opened or updated (There's a series of workflow runs from my commits on this that should be visible under the `Checks` tab and there is drop down next to the commit sha that has the results of all the workflows from the initial creation of the PR).

As for the yaml file name, I'm amenable to anything that makes sense.  At the tail end of this, it is doing a code coverage run which requires the unit tests to be built but the naming conventions I've seen online are all over the place (some even just use `test.yaml` to do all testing/coverage/linting).

peverwhee: ah yes, that makes sense.

in terms of the workflow name, i'm fine with `test.yaml` assuming that we're going to keep everything in one workflow rather than split them up like in CAM-SIMA

mwaxmonsky:
> couple small questions.
> 
> Also, I know you've documented this stuff on the new doc page (thanks!), but can I also request a README in the test/unit-test directory that describes, at minimum, how to run the tests?

Added a README with a few instructions and pointer to the gh-docs page but just let me know if there's any other details you think we should add.

test/unit-test/README.md
```diff
@@ -0,0 +1,18 @@
+# Unit Tests
+
+To add or update unit tests, follow the instructions from the [development guide](https://escomp.github.io/CAM-SIMA-docs/atmospheric_physics/development_workflow/#5-unit-testing).  Also, make sure [pFUnit](https://github.com/Goddard-Fortran-Ecosystem/pFUnit) is built and installed following the [build directions](https://github.com/Goddard-Fortran-Ecosystem/pFUnit?tab=readme-ov-file#building-and-installing-pfunit) (see the associated workflow file for a detailed example).
+
+To run the tests, from the root directory of your clone, run:
+
```
peverwhee, line 6: Can we include instructions on building pFUnit?
```
git clone --depth 1 --branch v4.10.0 https://github.com/Goddard-Fortran-Ecosystem/pFUnit.git
cd pFUnit
pwd
cmake -B./build -S.
cd build
make install
```

peverwhee (approved):
thanks! one request on the new readme. and reminder to pick a new name for the workflow yaml (assuming you're ok with changing it)

test/unit-test/README.md
mwaxmonsky: Added!

There's no mechanism to inject text from another file at the moment but I might be able to make a small shell script with this and then the docs can expand that in place and be used by the github workflow but that's a change for another day.

peverwhee: also, not to harp on this, but the steps to run the test are explicitly listed here even though they're also in the workflow. Is it that much worse to include the explicit instructions for how to build pFUnit here?

mwaxmonsky: Added a link to the workflow file!

I think we want to keep our duplication of instructions to a minimum as (at least in this case), this script is going to change and there's a risk they will diverge at some point.

It's not going to be difficult to fix or triage if we notice something incorrect later, it's just an additional location of documentation we need to be aware of as we make changes going forward.

.github/workflows/code-coverage.yaml
mwaxmonsky: So after a little more digging, `test.yaml` might be a little too generic (as there's other tests we'll want to add down the line aside from unit tests) so I went with `unit-tests.yaml` at the moment but if you think it should be more explicit or generic, happy to change it!

PR #147, opened 2024-10-31
Replace dynamic constituent routine with register phase

peverwhee:
Originator(s): peverwhee

Summary (include the keyword ['closes', 'fixes', 'resolves'] and issue number): Modifies existing dynamic constituent handling for the MUSICA interface to use the new register phase

(Helpful git command: git diff --name-status development...<your_branch_name>)

M schemes/musica/musica_ccpp.F90
M schemes/musica/muscia_ccpp.meta

- remove dynamic_constituent_routine from metadata
- add necessary comment header to Fortran
- add metadata for register routine

List any test failures:

Is this a science-changing update? New physics package, algorithm change, tuning changes, etc? No

### DIFF
--- a/schemes/musica/musica_ccpp.meta
+++ b/schemes/musica/musica_ccpp.meta
@@ -2,8 +2,29 @@
   name = musica_ccpp
   type = scheme
   dependencies = micm/musica_ccpp_micm.F90,micm/musica_ccpp_micm_util.F90,tuvx/musica_ccpp_tuvx.F90,tuvx/musica_ccpp_tuvx_height_grid.F90,musica_ccpp_util.F90
-  dynamic_constituent_routine = musica_ccpp_register
 
+[ccpp-arg-table]
+  name  = musica_ccpp_register
+  type  = scheme
+[ solver_type ]
+  standard_name = micm_solver_type
+  units = none
+  type = integer
+  dimensions = ()
+  intent = in
+[ num_grid_cells ]
+  standard_name = number_of_grid_cells
+  units = count
+  type = integer
+  dimensions = ()
+  intent = in
+[ constituent_props ]
+  standard_name = dynamic_constituents_for_musica_ccpp
+  units = none
+  dimensions = (:)
+  allocatable = True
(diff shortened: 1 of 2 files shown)


schemes/musica/musica_ccpp.meta
```diff
@@ -2,7 +2,41 @@
...
+  dimensions = ()
+  intent = in
+[ num_grid_cells ]
+  standard_name = number_of_grid_cells
+  units = count
+  type = integer
+  dimensions = ()
+  intent = in
+[ constituent_props ]
+  standard_name = dynamic_constituents_for_musica_ccpp
```
jimmielin, line 22: Thanks @peverwhee! I'm not totally familiar with the dynamic constituents so this may be a silly question - I understand this is the constituent properties object itself instead of the pointer (standard name `ccpp_constituent_properties`) but can I ask if there is a reason why the standard name for this is a musica_ccpp specific name rather than something like `ccpp_constituent_properties_object`?

jimmielin (approved):

schemes/musica/musica_ccpp.meta
peverwhee: Good question! There's a lot of constituent object updates swirling around. This is not *the* constituent properties object (the one controlled by the framework), but rather a smaller constituent properties object that contains any runtime constituents specifically for musica_ccpp. The standard name for this doesn't really matter as long as it's unique. Does that make more sense?

I'm working on updating the CAM-SIMA developer documentation to include usage of this new constituent usage in the register phase

jimmielin: Thank you @peverwhee for the explanation! This is very useful to know, looking forward to learning more about this in the CAM-SIMA developer documentation.

peverwhee: @jimmielin the new documentation is [here](https://escomp.github.io/CAM-SIMA-docs/design/constituents/#constituent-usage) if you're interested. If it's still confusing, let me know and I can document it better!

PR #167, opened 2024-11-22
Develop 100 cloud optics

### LINKED ISSUE #100: Update cloud optical properties in TUV-x before calculating rate constants
Update the wavelength-resolved optical depth, single scattering albedo, and asymmetry parameter for clouds in TUV-x prior to calculating photolysis rate constants

**Acceptance Criteria**
- TUV-x photolysis rate constants are calculated for the actual heights for each column at each time step

**Ideas**
- Identify the CCPP standard names for cloud fraction and liquid water content
- Follow the logic of the TUV-x PR into CAM: https://github.com/ESCOMP/CAM/blob/c12d1e46e0fdc1dccb0a651a6c9fefd6bb80b2ba/src/chemistry/mozart/mo_tuvx.F90#L1631-L1652

mattldawson:
Adds the calculation of cloud optical properties to TUV-x for photolysis rate constant calculations. The algorithm is adapted from a post-calculation modification of photolysis rate constants calculated by the lookup version of TUV-x in CAM to account for the effects of clouds, [here](https://github.com/ESCOMP/CAM/blob/47dd341f83afc1e1faf28fd4cb6e463629db4c19/src/chemistry/mozart/mo_photo.F90#L949-L965)

closes #100


schemes/musica/tuvx/musica_ccpp_tuvx_cloud_optics.F90
```diff
@@ -0,0 +1,128 @@
...
+
+  private
+  public :: create_cloud_optics_radiator, set_cloud_optics_values
+
+  ! This module is used to set the optical properties of clouds in TUV-x.
+  ! Optical properties are defined as a function of wavelength and height,
+  ! and include the cloud optical depth, single scattering albedo,
+  ! and asymmetry parameter.
+  !
+  ! See module_ccpp_tuvx_height_grid for the definition of the height grid
```
boulderdaze, line 14: Probably 'musica_ccpp_tuvx_height_grid'?

mattldawson: ah, yes - fixed!

schemes/musica/tuvx/musica_ccpp_tuvx_cloud_optics.F90
```diff
@@ -0,0 +1,128 @@
...
+      errmsg, errcode ) result( radiator )
+    use musica_ccpp_util,     only: has_error_occurred
+    use musica_tuvx_grid,     only: grid_t
+    use musica_tuvx_radiator, only: radiator_t
+    use musica_util,          only: error_t
+
+    type(grid_t),                        intent(inout) :: height_grid
+    type(grid_t),                        intent(inout) :: wavelength_grid
+    character(len=*),                    intent(out)   :: errmsg
+    integer,                             intent(out)   :: errcode
```
boulderdaze, line 41: Is the large spacing intentional?

mattldawson: not by me. copilot may have had a reason, but I've removed the extra spaces.

test/musica/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ target_sources(test_musica_api
     ${CCPP_SRC_PATH}/ccpp_hash_table.F90
     ${CCPP_SRC_PATH}/ccpp_hashable.F90
     ${CCPP_SRC_PATH}/ccpp_types.F90
+    ${TO_BE_CCPPIZED_SRC_PATH}/ccpp_const_utils.F90
```
boulderdaze, line 36: I might suggest adding it after line 31, just because they come from the same sources, but it's not a big deal.

schemes/musica/musica_ccpp.F90
```diff
@@ -61,7 +73,8 @@ subroutine musica_ccpp_run(time_step, temperature, pressure, dry_air_density, co
                              surface_temperature, surface_albedo,                                    &
                              number_of_photolysis_wavelength_grid_sections,                          &
                              photolysis_wavelength_grid_interfaces, extraterrestrial_flux, &
-                             standard_gravitational_acceleration, errmsg, errcode)
+                             standard_gravitational_acceleration, cloud_area_fraction,             &
+                             air_pressure_thickness, errmsg, errcode)
```
boulderdaze, line 77: Not a big deal, but maybe we could match the spacing around the `&`?

boulderdaze (changes requested):
Looks good! I only have a couple of minor suggestions

schemes/musica/tuvx/musica_ccpp_tuvx_cloud_optics.F90
```diff
@@ -0,0 +1,128 @@
...
+                                      reciprocal_of_gravitational_acceleration, &
+                                      errmsg, errcode )
+    use ccpp_kinds,           only: kind_phys
+    use musica_ccpp_util,     only: has_error_occurred
+    use musica_tuvx_radiator, only: radiator_t
+    use musica_util,          only: error_t
+
+    type(radiator_t), intent(inout) :: radiator
+    real(kind_phys),  intent(in)    :: delta_pressure(:) ! pressure delta about vertical level midpoints (Pa)
+    real(kind_phys),  intent(in)    :: cloud_fraction(:) ! (unitless)
```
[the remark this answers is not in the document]
mattldawson: thanks for catching this! fixed!

schemes/musica/tuvx/musica_ccpp_tuvx_cloud_optics.F90
```diff
@@ -0,0 +1,128 @@
...
+    real(kind_phys),  intent(in)    :: cloud_liquid_water_content(:) ! (kg/kg)
+    real(kind_phys),  intent(in)    :: reciprocal_of_gravitational_acceleration ! (s^2/m)
+    character(len=*), intent(out)   :: errmsg
+    integer,          intent(out)   :: errcode
+
+    ! local variables
+    type(error_t)   :: error
+    real(kind_phys) :: od(num_vertical_levels)
```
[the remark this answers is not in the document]
mattldawson: good call. I've renamed it as `cloud_fraction`

schemes/musica/tuvx/musica_ccpp_tuvx_cloud_optics.F90
```diff
@@ -0,0 +1,128 @@
...
+    if ( size(cloud_liquid_water_content) /= n_host_midpoints ) then
+      errmsg = "[MUSICA Error] Invalid size of cloud liquid water content for TUV-x."
+      errcode = 1
+      return
+    end if
+    if ( n_host_midpoints + 1 /= num_vertical_levels ) then
+      errmsg = "[MUSICA Error] Invalid size of cloud fraction for TUV-x."
+      errcode = 1
+      return
+    end if
```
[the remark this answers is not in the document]
mattldawson: the last block is what was checking that the size of `cloud_fraction` is correct, but maybe the `n_host_midpoints` was making this unclear. I've removed the `n_host_midpoints` and replaced it with `size(cloud_fraction)`. Let me know if this looks better.

test/musica/CMakeLists.txt
mattldawson: moved!

schemes/musica/musica_ccpp.F90
mattldawson: fixed!

schemes/musica/tuvx/musica_ccpp_tuvx_cloud_optics.F90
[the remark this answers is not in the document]
boulderdaze: I see. Do you think it makes sense to check the cloud fraction size first, since the other two arrays are attempting to compare sizes under the assumption that the cloud fraction size is correct? Also, since the `size` function called five times, I think it might be better to define the variables as in the previous version.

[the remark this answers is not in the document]
mattldawson: Good idea. updated!

boulderdaze (approved):
Looks good! Thank you for addressing the comments!

schemes/musica/musica_ccpp.F90
```diff
@@ -23,28 +23,40 @@ subroutine musica_ccpp_register(solver_type, num_grid_cells, constituent_props,
     character(len=512),                               intent(out) :: errmsg
     integer,                                          intent(out) :: errcode
 
-    call micm_register(solver_type, num_grid_cells, constituent_props, errmsg, errcode)
+    type(ccpp_constituent_properties_t), allocatable :: constituent_set(:)
+
+    call micm_register(solver_type, num_grid_cells, constituent_set, errmsg, errcode)
+    if (errcode /= 0) return
+    constituent_props = constituent_set
```
boulderdaze, line 30: I branched off from yours to create a new one and have a couple of questions. I think this line is unnecessary because when MICM fails, it returns empty, and when TUVX fails, it can only return MICM constituents, but I think CAM-SIMA should abort the process. What do you think?

schemes/musica/musica_ccpp.F90
```diff
@@ -23,28 +23,40 @@ subroutine musica_ccpp_register(solver_type, num_grid_cells, constituent_props,
...
+
+    call tuvx_register(constituent_set, errmsg, errcode)
+    if (errcode /= 0) return
+    constituent_props = [ constituent_props, constituent_set ]
```
boulderdaze, line 34: Another thought is that these two variable names `constituent_props` and `constituent_set` don't seem informative. What does 'set' mean differently from 'props'?

schemes/musica/musica_ccpp.F90
mattldawson: I think line 29 should return a non-zero error code from this function if the MICM register function fails, and line 33 should do the same if TUV-x fails. Line 30 just copies the constituents needed by MICM to an array that will hold the cumulative set of constituents needed for both MICM and TUV-x, so that `constituent_set` can be reused in line 32 to collect the TUV-x constituents.

schemes/musica/musica_ccpp.F90
mattldawson: Good point. These weren't good name choices. Maybe `constituent_props_subset` instead? to indicate that it is just used to temporarily hold the specific species needed for MICM, and later TUV-x?

schemes/musica/musica_ccpp.F90
boulderdaze: Ah, thank you, that makes sense. Just curious—don't we need a deallocation call for the allocated array `constituent_set` before passing it into the TUVX register function? I'm wondering because the size of the array should be different when it is passed in to MICM and TUV as an argument.

schemes/musica/musica_ccpp.F90
boulderdaze: Yeah, that sounds good to me. Thank you!

schemes/musica/musica_ccpp.F90
mattldawson: the `constituent_props` argument of the `tuvx_register()` function is `intent(out)` and in fortran allocatable variables are automatically deallocated when they go out of scope, so this _should_ be ok, but I will add a `deallocate()` just in case (you never know with fortran compilers...)

boulderdaze: oh I missed the `intent(out)`. My bad, thank you for clarifying it!

to_be_ccppized/ccpp_const_utils.F90
```diff
@@ -13,13 +13,13 @@ subroutine ccpp_const_get_idx(constituent_props, name, cindex, errmsg, errflg)
     use ccpp_constituent_prop_mod, only: ccpp_constituent_prop_ptr_t
 
     ! Input arguments
-    type(ccpp_constituent_prop_ptr_t), pointer, intent(in)  :: constituent_props(:)
-    character(len=*),                           intent(in)  :: name   ! constituent name
+    type(ccpp_constituent_prop_ptr_t), intent(in)  :: constituent_props(:)
```
jimmielin, line 16: Could I ask if there's any particular reason the `pointer` attribute was removed from here?

`ccpp_const_get_idx` is used in `cam_constituents` as well so this change would affect existing code - could you confirm that CAM-SIMA still passes tests after removing the `pointer` attribute here? I think it is _fine_ if the actual argument is a `pointer` (in `cam_constituents::const_props`) but the dummy argument is not, as long as it's read only, but it would be good to check...

jimmielin (changes requested):
Thanks @mattldawson! I have a couple questions regarding the use of `ccpp_const_utils` but otherwise the code looks great.

schemes/musica/tuvx/musica_ccpp_tuvx.F90
```diff
@@ -106,6 +159,11 @@ subroutine tuvx_init(vertical_layer_dimension, vertical_interface_dimension, &
     type(mappings_t),      pointer :: photolysis_rate_constants_ordering
     type(error_t)                  :: error
 
+    ! Get needed indices in constituents array
+    call ccpp_const_get_idx(constituent_props, CLOUD_LIQUID_WATER_CONTENT_LABEL, &
+                            index_cloud_liquid_water_content, errmsg, errcode)
+    if (errcode /= 0) return
```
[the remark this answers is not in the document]
mattldawson: Hi @jimmielin - thanks for catching this! I added a check to make sure the index was found.

to_be_ccppized/ccpp_const_utils.F90
mattldawson: Thanks for pointing this out, I had meant to add a comment about this, but I forgot when I sent submitted the PR.

I removed the `pointer` attribute because we don't include this attribute for `constituent_props` in the musica module functions, and it isn't actually needed in the `ccpp_const_get_idx` function, as you point out.

However, if this function is being used through an interface (e.g., for a function pointer or an implementation of a deferred type-bound procedure) where the `pointer` attribute couldn't be removed for some reason, we could do a separate implementation of the `ccpp_const_get_idx` function for non-pointers. Given the intent of this function, it seemed pretty unlikely to me that this was the case, but I agree it's good to check. Maybe @peverwhee could weigh in?

jimmielin: Thanks @mattldawson! If it isn't actually needed (should be fine if it's `intent(in)`) I'm happy to accept the change, just wanted to make sure it wouldn't break anything else. I'll leave this conversation open in case @peverwhee would like to add comments but I'll approve the PR now - no need to wait for me!

jimmielin (approved):

to_be_ccppized/ccpp_const_utils.F90
peverwhee: i'm ok with removing the pointer attribute as well - I just ran the regression tests with it removed and they work fine so I'll agree with @jimmielin that it isn't needed.

mattldawson: awesome, thanks!

PR #171, opened 2024-11-27
Solar zenith angle and Earth-Sun distance

### LINKED ISSUE #163: Set solar zenith angle in TUV-x
Set the solar zenith angle in TUV-x prior to calculating photolysis rate constants.

**Acceptance Criteria**
- Solar zenith angle is set based on model state before photolysis rate constants are calculated in TUV-x

**Ideas**
- create a CCPP scheme for calculating solar zenith angles based on [CAM zenith function](https://github.com/ESCOMP/CAM/blob/47dd341f83afc1e1faf28fd4cb6e463629db4c19/src/utils/orbit.F90#L5-L64)
- alternatively move CAM function to `to_be_ccppized` folder
- convert solar zenith angle to degrees from cos(sza)

### LINKED ISSUE #164: Set Earth-Sun distance in TUV-x
Set the Earth-Sun distance in TUV-x before calculating photolysis rate constants.

**Acceptance Criteria**
- Earth-Sun distance is set in TUV-x based on model state prior to calculating photolysis rate constants

**Ideas**
- Follow [logic used in CAM](https://github.com/ESCOMP/CAM/blob/47dd341f83afc1e1faf28fd4cb6e463629db4c19/src/chemistry/mozart/mo_gas_phase_chemdr.F90#L810-L812) to calculate Earth-Sun distance
- Identify how to access modules like `shr_orb_mod` not present in CAM codebase

mattldawson:
Adds the calculation of solar zenith angle and Earth-Sun distance needed as inputs to TUV-x.

closes #163 
closes #164 

Both calculations required functions that are currently in shared code in CAM [here](https://github.com/ESCOMP/CESM_share/blob/f6f31fd61cb8f80aee97311fcca64b3e26b0202c/src/shr_orb_mod.F90#L736-L814) and [here](https://github.com/ESCOMP/CESM_share/blob/f6f31fd61cb8f80aee97311fcca64b3e26b0202c/src/shr_orb_mod.F90#L110-L159) that I put in a modified form in the `to_be_ccppized/` folder. @nusbaume @peverwhee - if there is a better way to handle these functions, let me know and I can update the PR.

Requires an update to CAM-SIMA that is in review to make certain variables available through CCPP (https://github.com/ESCOMP/CAM-SIMA/pull/325)

After more discussion, the plan is to use the shared functions in CAM-SIMA and pass the outputs through CCPP. So, this PR will be put in draft until this issue is finished: https://github.com/ESCOMP/CAM-SIMA/issues/328

A related draft PR to CAM-SIMA has been added (https://github.com/ESCOMP/CAM-SIMA/pull/337) and will be un-drafted once this PR is merged in.


schemes/musica/musica_ccpp_util.F90
```diff
@@ -2,10 +2,15 @@
...
 
+  use ccpp_kinds, only: kind_phys
+
   implicit none
 
   private
-  public :: has_error_occurred
+  public :: has_error_occurred, calculate_solar_zenith_angle_and_earth_sun_distance
+
+  real(kind_phys), parameter, public :: PI = 3.14159265358979323846_kind_phys
```
jimmielin, line 12: I would suggest avoid hardcoding this in a separate location and use the CCPP standard name `pi_constant` if possible, or if not, maybe include from `shr_const_pi` (these are the same numerically). I realize this is used in another module and maybe the CCPP variable can be brought in from there.

schemes/musica/tuvx/musica_ccpp_tuvx.F90
```diff
@@ -129,6 +132,7 @@ subroutine tuvx_init(vertical_layer_dimension, vertical_interface_dimension, &
     use musica_tuvx, only: grid_map_t, profile_map_t, radiator_map_t
     use musica_util, only: error_t, configuration_t
     use musica_ccpp_namelist, only: filename_of_tuvx_micm_mapping_configuration
+    use musica_ccpp_util, only: PI
```
jimmielin, line 135: Maybe use CCPP variable `pi_constant` as an argument, if possible?

schemes/musica/tuvx/musica_ccpp_tuvx.F90
```diff
@@ -415,13 +419,14 @@ subroutine tuvx_run(temperature, dry_air_density,                  &
...
                       standard_gravitational_acceleration,           &
+                      solar_zenith_angle, earth_sun_distance,        &
                       cloud_area_fraction, constituents,             &
                       air_pressure_thickness, rate_parameters,       &
                       errmsg, errcode)
     use musica_util,                               only: error_t
     use musica_ccpp_tuvx_height_grid,              only: set_height_grid_values, calculate_heights
     use musica_ccpp_tuvx_temperature,              only: set_temperature_values
-    use musica_ccpp_util,                          only: has_error_occurred
+    use musica_ccpp_util,                          only: has_error_occurred, PI
```
jimmielin, line 429: Similarly, would it be possible to use the CCPP `pi_constant` here as well?

to_be_ccppized/shr_orb_mod.F90
```diff
@@ -0,0 +1,141 @@
+! Copyright (C) 2024 National Science Foundation-National Center for Atmospheric Research
+! SPDX-License-Identifier: Apache-2.0
+module shr_orb_mod
```
jimmielin, line 3: I suppose this is fine since it's a temporary solution but I wonder if the copy of `shr_orb_mod` in CAM-SIMA's `share` submodule can be used directly here, perhaps through a dependency (`dependencies = ../../../../../share/shr_orb_mod.F90` in the ccpp metadata. Just wanted to open it up for discussion in case it's an easy fix. Please don't hold up the PR if not, I think this is best computed in the host model and passed as a CCPP variable eventually (e.g., `cosine_of_solar_zenith_angle`)

jimmielin (approved):
Thanks @mattldawson! I have a couple of minor suggestions but nothing that holds up the PR. Just a reminder that the Earth eccentricity, etc. parameter need a SIMA update (https://github.com/ESCOMP/CAM-SIMA/pull/325) to be merged before they're available for use by code in this PR.

mattldawson:
> Thanks @mattldawson! I have a couple of minor suggestions but nothing that holds up the PR. Just a reminder that the Earth eccentricity, etc. parameter need a SIMA update ([ESCOMP/CAM-SIMA#325](https://github.com/ESCOMP/CAM-SIMA/pull/325)) to be merged before they're available for use by code in this PR.

Good point. I added a note to the PR description. thanks!

schemes/musica/musica_ccpp_util.F90
mattldawson: I guess my preference would be to use the `shr_const_pi` instead of adding yet another function argument to a list that is getting pretty long, and then having to pass constants down through all the internal functions, but I'll go with the approach the group prefers.

schemes/musica/musica_ccpp_util.F90
```diff
@@ -37,4 +42,53 @@ logical function has_error_occurred(error, error_message, error_code)
...
+    real(kind_phys) :: delta
+    integer :: i_sza
+
+    errcode = 0
+    errmsg = ''
+
+    ! Calculate earth/orbit parameters
+    call shr_orb_decl(calendar_day, earth_eccentricity, earth_obliquity, &
+                      perihelion_longitude, moving_vernal_equinox_longitude, &
+                      delta, earth_sun_distance)
```
boulderdaze, line 85: This function is imported, but if we or CAM-SIMA are going to rewrite it, is there a chance we could rename the function?

schemes/musica/tuvx/musica_ccpp_tuvx.F90
```diff
@@ -469,18 +475,25 @@ subroutine tuvx_run(temperature, dry_air_density,                  &
...
-      if (errcode /= 0) return
+        if (errcode /= 0) return
 
-      call set_temperature_values( temperature_profile, temperature(i_col,:), &
+        call set_temperature_values( temperature_profile, temperature(i_col,:), &
                                    surface_temperature(i_col), errmsg, errcode )
-      if (errcode /= 0) return
+        if (errcode /= 0) return
 
       call set_cloud_optics_values( cloud_optics, cloud_area_fraction(i_col,:), &
```
boulderdaze, line 498: Could you add space in the lines starting 498 - 503 to align with the `else` conditional code?

test/musica/test_musica_api.F90
```diff
@@ -317,7 +336,9 @@ subroutine test_chapman()
                           geopotential_height_wrt_surface_at_interface, surface_geopotential,           &
                           surface_temperature, surface_albedo, num_photolysis_wavelength_grid_sections, &
                           flux_data_photolysis_wavelength_interfaces, extraterrestrial_flux,            &
-                          standard_gravitational_acceleration, cloud_area_fraction, air_pressure_thickness, errmsg, errcode )
+                          standard_gravitational_acceleration, cloud_area_fraction, air_pressure_thickness, latitude, longitude, earth_eccentricity, &
+                          earth_obliquity, perihelion_longitude, moving_vernal_equinox_longitude,       &
+                          calendar_day, errmsg, errcode )
```
boulderdaze, line 341: Would it be possible to split this list of arguments (line 339) into two lines?

test/musica/test_musica_api.F90
```diff
@@ -443,6 +475,26 @@ subroutine test_terminator()
...
+    ! Set conditions for one daytime and one nighttime column
+    ! Greenwich, UK and Wellington, NZ
+    latitude = (/ 51.5_kind_phys, -41.3_kind_phys /)
+    longitude = (/ 0.0_kind_phys, 174.8_kind_phys /)
+    earth_eccentricity = 0.0167_kind_phys
+    earth_obliquity = 23.5_kind_phys * DEGREE_TO_RADIAN
+    perihelion_longitude = 102.9_kind_phys * DEGREE_TO_RADIAN
+    moving_vernal_equinox_longitude = 182.7_kind_phys * DEGREE_TO_RADIAN
+    calendar_day = 183.5_kind_phys ! noon GMT Jul 1
+
```
boulderdaze, line 497: These are duplicates (lines 478–486). Could you remove them?

test/musica/test_musica_api.F90
```diff
@@ -533,7 +585,9 @@ subroutine test_terminator()
                           geopotential_height_wrt_surface_at_interface, surface_geopotential,           &
                           surface_temperature, surface_albedo, num_photolysis_wavelength_grid_sections, &
                           flux_data_photolysis_wavelength_interfaces, extraterrestrial_flux,            &
-                          standard_gravitational_acceleration, cloud_area_fraction, air_pressure_thickness, errmsg, errcode )
+                          standard_gravitational_acceleration, cloud_area_fraction, air_pressure_thickness, latitude, longitude, earth_eccentricity, &
+                          earth_obliquity, perihelion_longitude, moving_vernal_equinox_longitude,       &
```
boulderdaze, line 589: The same request for splitting the arguments into a couple of lines.

boulderdaze (changes requested):
Looks good! I just have a couple of suggestions and questions.

schemes/musica/musica_ccpp.F90
```diff
@@ -101,6 +104,14 @@ subroutine musica_ccpp_run(time_step, temperature, pressure, dry_air_density, co
     real(kind_phys),         intent(in)    :: standard_gravitational_acceleration               ! m s-2
     real(kind_phys),         intent(in)    :: cloud_area_fraction(:,:)                          ! unitless (column, level)
     real(kind_phys),         intent(in)    :: air_pressure_thickness(:,:)                       ! Pa (column, level)
+    real(kind_phys),         intent(in)    :: latitude(:)                                       ! radians (column)
+    real(kind_phys),         intent(in)    :: longitude(:)                                      ! radians (column)
+    
+    real(kind_phys),         intent(in)    :: earth_eccentricity                                ! Earth's eccentricity factor (unitless) (typically 0 to 0.1)
```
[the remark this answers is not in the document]
mattldawson: fixed!

schemes/musica/musica_ccpp.F90
```diff
@@ -121,6 +140,8 @@ subroutine musica_ccpp_run(time_step, temperature, pressure, dry_air_density, co
                   photolysis_wavelength_grid_interfaces,         &
                   extraterrestrial_flux,                         &
                   standard_gravitational_acceleration,           &
+                  solar_zenith_angle,                            &
+                  earth_sun_distance,                            &
```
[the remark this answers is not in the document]
mattldawson: good point, I've reordered the TUV-x run function arguments

schemes/musica/musica_ccpp_util.F90
mattldawson: I think the plan now is to use this function in CAM-SIMA and pass the inputs needed for TUV-x through CCPP, so I will remove this copy of these shared functions. (But, I totally agree with renaming this - it just might be that this function is used in a lot of places in CESM, so it might take some discussion first)

schemes/musica/tuvx/musica_ccpp_tuvx.F90
mattldawson: added!

test/musica/test_musica_api.F90
mattldawson: yes, updated

test/musica/test_musica_api.F90
mattldawson: thanks for catching this. I removed the duplicate

test/musica/test_musica_api.F90
mattldawson: done!

schemes/musica/musica_ccpp_namelist.xml
```diff
@@ -75,16 +75,20 @@
...
+    <group>musica_ccpp</group>
+    <standard_name>micm_solver_type</standard_name>
+    <units>none</units>
+    <desc>
+      The type of MICM solver to use.
+    </desc>
+    <values>
+      <value>1</value>
+    </values>
+  </entry>
```
boulderdaze, line 91: I’m hesitant about setting the solver type using an integer, as this implies that users need to know the enum type and which solver corresponds to each integer. Can I ask why we want to include the solver type in the xml?

test/musica/musica_ccpp_namelist.F90
```diff
@@ -4,7 +4,8 @@ module musica_ccpp_namelist
   implicit none
 
   private
-  
+
+  integer,            public :: micm_solver_type = 1
```
boulderdaze, line 8: I have the same concern about using an integer value to set the solver type.

boulderdaze (changes requested):
Looks good! I just have a couple minor questions.

schemes/musica/musica_ccpp.meta
```diff
@@ -147,23 +141,17 @@
...
-  dimensions = (horizontal_loop_extent)
+  units = m
+  dimensions = (photolysis_wavelength_grid_interface_dimension)
   intent = in
 [ extraterrestrial_flux ]
   standard_name = extraterrestrial_radiation_flux
   type = real | kind = kind_phys
-  units = photons cm-2 s-1 nm-1
-  dimensions = (horizontal_loop_extent)
+  units = count cm-2 s-1 nm-1
```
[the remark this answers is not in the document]
mattldawson: After talking with Michael yesterday it seemed like `photons` wouldn't be a valid unit and `count` was an existing unit that seemed appropriate, but I could be thinking about this wrong. @peverwhee or @nusbaume - do you have any recommendations on choosing units?

schemes/musica/musica_ccpp_namelist.xml
mattldawson: After I went to try to run the MUSICA scheme in CAM-SIMA, I realized that we have several inputs to MUSICA (like the solver type) that need to be provided from "somewhere". The options (at least that I'm aware of) were to put this in this xml file, or to make it available from CAM-SIMA, like the solar zenith angle, etc. I didn't seem like a good idea to me to put this in CAM-SIMA, so I put it here. But, maybe @nusbaume or @peverwhee might have a better suggestion for where to put this?

Also, I bet we could change this to a string and convert it to the enum in the fortran code. Would that be better do you think?

boulderdaze: I see. I think I would do the same and put the solver type in this xml file. I like the idea of using a string and converting it to an integer in the code, though I’m not sure if there’s a way to do this in a simple way in Fortran. But this explanation makes sense to me, and I don’t mind this implementation.

boulderdaze (approved):
Thank you for addressing the comments!

schemes/musica/musica_ccpp.meta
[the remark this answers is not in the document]
peverwhee: i'd vote for `cm-2 s-1 nm-1` (no count), but units aren't exactly my forte.

schemes/musica/musica_ccpp_namelist.xml
nusbaume: I think the location you have now (the namelist XML file) makes a lot of sense, so I would personally vote to just leave it there. Thanks!

schemes/musica/musica_ccpp.meta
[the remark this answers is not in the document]
nusbaume: I think @peverwhee's suggestion is best if we include the word `photon` somewhere in the standard name.  Otherwise I would probably choose `photons cm-2 s-1 nm-1` over `count cm-2 s-1 nm-1`, as `count` is usually reserved for dimension variables (e.g. `vertical_layer_dimension`).  Anyways, that's my two cents!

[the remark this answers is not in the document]
mattldawson: ah, ok - so is `photons cm-2 s-1 nm-1` a valid option? If so, I might lean in that direction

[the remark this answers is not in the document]
nusbaume: I don't see it listed in the dictionary yet, but I feel like that makes it clear as to what the "flux" actually is.  If for some reason the other CCPP folks object to it then we can always fix it later to whatever is agreed upon.

PR #174, opened 2024-12-06
Set gas-species profiles in TUV-x prior to calculating rate constants

boulderdaze:
Currently in draft for questions and discussions


schemes/musica/musica_ccpp.F90
```diff
@@ -111,20 +111,7 @@ subroutine musica_ccpp_run(time_step, temperature, pressure, dry_air_density, co
...
-                  surface_albedo,                                &
-                  number_of_photolysis_wavelength_grid_sections, &
-                  photolysis_wavelength_grid_interfaces,         &
-                  extraterrestrial_flux,                         &
-                  standard_gravitational_acceleration,           &
-                  cloud_area_fraction, constituents,             &
-                  air_pressure_thickness, rate_parameters,       &
-                  errmsg, errcode)
-
+    ! TODO(jiwon) this might not be correct because it doesn't know the index
```
boulderdaze, line 114: Hi @peverwhee, is this code still valid? We discussed that the sequence of the constituents array is not guaranteed. Do we now need to search for the index of the MICM constituents before attempting to retrieve the molar mass?
```fortran
    do i_elem = 1, size(molar_mass_arr)
      call constituent_props(i_elem)%molar_mass(molar_mass_arr(i_elem), errcode, errmsg)
      if (errcode /= 0) then
        errmsg = "[MUSICA Error] Unable to get molar mass."
        return
      end if
    end do
```

schemes/musica/tuvx/musica_ccpp_tuvx_gas_species_profile.F90
```diff
@@ -0,0 +1,219 @@
...
+  public :: create_gas_species, configure_gas_species, create_gas_species_profile_group, &
+            set_gas_species_values, deallocate_gas_species, deallocate_gas_species_profile_group
+
+  !> Conversion factor from km to cm
+  real(kind_phys), parameter, public :: km_to_cm = 1.0e5
+  !> Conversion factor from m3 to cm3
+  real(kind_phys), parameter, public :: m_3_to_cm_3 = 1.0e6
+  !> Molar mass value of dry air is obtained from 'CAM-SIMA/src/utils/std_atm_profile.F90'
+  ! TODO(jiwon)
+  real(kind_phys), parameter, public :: MOLAR_MASS_DRY_AIR = 0.0289644_kind_phys ! kg mol-1
```
boulderdaze, line 17: Hi @mattldawson and @nusbaume,  I get the molar mass value of dry air from [CAM-SIMA/src/utils/std_atm_profile.F90](https://github.com/ESCOMP/CAM-SIMA/blob/development/src/utils/std_atm_profile.F90#L48). Is it okay to define it here?  We discussed about passing parameters from CAM-SIMA, but I'm still unsure about the approach we're going to take.

schemes/musica/tuvx/musica_ccpp_tuvx_gas_species_profile.F90
```diff
@@ -0,0 +1,219 @@
...
+  
+    ! Air
+    ! Find the index of the species and create a 'gas_species_t' for it
+    ! call ccpp_const_get_idx(constituent_props, "air", index_air, errmsg, errcode)
+    ! write(*,*) "air index: ", index_air
+    ! if (errcode /= 0) return
+
+    ! call constituent_props(index_air)%molar_mass(molar_mass_air, errcode, errmsg)
+    ! if (errcode /= 0) return
+
```
boulderdaze, line 91: Hi @mattldawson, I was trying to make the gas species configurable, allowing the user to provide their own species and define a function. Do you have any feedback on this implementation? Also, `air` is not included in the `species.json`, so it ends up being hard-coded. Could you share your thoughts on that in my comment?

schemes/musica/tuvx/musica_ccpp_tuvx_gas_species_profile.F90
nusbaume: Hi @boulderdaze, you should be able to read in the molar mass of dry air as an input argument with the standard name `molecular_weight_of_dry_air`, which will be in units of `g mol-1`.  There is also a composition-dependent version as well, but that is only used in WACCM and WACCM-X configurations (although if you would like to use that quantity instead just let me know).

The reason we generally suggest passing this value in as an argument is because some host models and host model configurations might have slightly different values for the dry air molar mass, especially if they are configured for high-top, paleoclimate, or exoplanet simulations, and so by bringing it in as an argument we won't have to worry about molar mass value mismatches.

Anyways, hopefully that helps?  If for some reason that method does not work for your particular setup here just let me know.  Thanks!

boulderdaze: Ah that makes sense, thank you for the additional details! I will use `molecular_weight_of_dry_air` as an input.

schemes/musica/musica_ccpp.F90
peverwhee: hi @boulderdaze sorry for the delay!

Yes, you're correct that the order is not guaranteed. So you'll have to grab the index by the standard name. Haipeng created `ccpp_const_get_idx` in https://github.com/ESCOMP/atmospheric_physics/blob/main/to_be_ccppized/ccpp_const_utils.F90 for this purpose.

boulderdaze: Great, thank you for confirming!

boulderdaze:
Closes this issue and reopens #184

Issue #165, opened 2024-11-22
Map indices between constituents and MICM


mattldawson:
There is no guarantee that the constituents array will be ordered as specified in the `musica_ccpp_micm::micm_register()` function because other schemes will be adding constituents as well. Ensure that the constituent array is properly mapped to the MICM state prior to solving chemistry.

**Acceptance Criteria**
- Constituents are properly mapped to MICM state, even when not in the order specified during the `micm_register()` function

**Ideas**
- Create a map of indices in a `micm_init()` function between MICM state and constituents using the `micm%species_ordering` array and the [CCPP constituent index lookup function](https://github.com/ESCOMP/atmospheric_physics/blob/d755e195961a155730371009dd84c9b104b0ddcd/to_be_ccppized/ccpp_const_utils.F90#L12-L48)
- Possibly add a function to `ccpp_const_utils` that returns an array of constituent indices based on an array of names

PR #184, opened 2024-12-23
Set gas-species profiles in TUV-x and map indices between constituents and MICM

### LINKED ISSUE #98: Set gas-species profiles in TUV-x prior to calculating rate constants
Update the concentrations of radiatively active gas-phase species in TUV-x prior to calculating photolysis rate constants.

@peverwhee - TUV-x needs to know the concentrations of certain chemical species above the top of the model. Is there a way to set this value for a constituent?

(related issue on CCPP-framework: https://github.com/NCAR/ccpp-framework/issues/573)

**Acceptance Criteria**
- Gas-phase species can be configured to be included in the TUV-x solving for the radiation field.
- The concentrations above the model top can be configured to be estimated using a scale height, or read in from a file
- TUV-x photolysis rate constants are calculated using the actual concentrations of total air density and each specified gas-phase species for each column at each timestep

**Ideas**
- Identify CCPP standard name for air density
- Use `constituents_t` type to update species concentrations
- Follow the logic of the TUV-x PR into CAM: https://github.com/ESCOMP/CAM/blob/c12d1e46e0fdc1dccb0a651a6c9fefd6bb80b2ba/src/chemistry/mozart/mo_tuvx.F90#L1508-L1619

**Questions**
- Can we enable users to create `profile` and set the values for them?
- The indexing of the constituents is not guaranteed to be the same as what we set in `micm_register()` .  We might need to add an initialization step where we look up the indices of chemical species (fixed and variable)

boulderdaze:
Closes #98 and #165 
Currently in draft

### DIFF
--- a/schemes/musica/musica_ccpp_species.F90
+++ b/schemes/musica/musica_ccpp_species.F90
@@ -1,6 +1,4 @@
 module musica_ccpp_species
-  ! This module owns musica species
-
   use ccpp_kinds, only: kind_phys
 
   implicit none
@@ -26,10 +24,6 @@ module musica_ccpp_species
   interface musica_species_t
     procedure species_t_constructor
   end interface musica_species_t
-
-  type, public :: musica_species_ptr_t
-    type(musica_species_t), pointer :: species
-  end type musica_species_ptr_t
 
   ! Species are ordered to match the sequence of the MICM state array
   type(musica_species_t), allocatable, protected, public :: micm_species_set(:) ! index should match with the MICM state array
--- a/schemes/musica/musica_ccpp_species.F90
+++ b/schemes/musica/musica_ccpp_species.F90
@@ -1,0 +1,296 @@
+module musica_ccpp_species
+  ! This module owns musica species
+
+  use ccpp_kinds, only: kind_phys
+
+  implicit none
+  private
+
+  public :: cleanup_musica_species, register_musica_species, initialize_musica_species_indices, &
+            initialize_molar_mass_array, extract_subset_constituents, update_constituents
+
+  integer, parameter, public :: MUSICA_INT_UNASSIGNED = -99999
+
+  ! Species are ordered to match the sequence of the MICM state array
+  type(musica_species_t), allocatable, protected, public :: micm_species_set(:) ! index should match with the MICM state array
+  type(musica_species_t), allocatable, protected, public :: tuvx_species_set(:)
+  integer,                allocatable, protected, public :: micm_indices_constituent_props(:)
+  integer,                allocatable, protected, public :: tuvx_indices_constituent_props(:)
+  real(kind_phys),        allocatable, protected, public :: micm_molar_mass_array(:) ! kg mol-1
+  integer,                             protected, public :: number_of_micm_species = MUSICA_INT_UNASSIGNED
+  integer,                             protected, public :: number_of_tuvx_speceis = MUSICA_INT_UNASSIGNED
+
+  !> Definition of the gas species type
+  type, public :: musica_species_t
+    character(len=:), allocatable :: name
+    character(len=:), allocatable :: unit
+    real(kind_phys)               :: molar_mass ! kg mol-1
+    integer                       :: index_musica_species = MUSICA_INT_UNASSIGNED
+    integer                       :: index_constituent_props = MUSICA_INT_UNASSIGNED
+    logical                       :: profiled = .false. ! optional
+    real(kind_phys)               :: scale_height = 0.0_kind_phys ! km, optional
+  end type musica_species_t
+
+  type, public :: musica_species_ptr_t
+    type(musica_species_t), pointer :: species
+  end type musica_species_ptr_t
+
+contains
+
+  !> Constructor for gas_species_t object
+  function species_t_constructor(name, molar_mass, index_musica_species, &
+    index_constituent_props) result( this )
+
+    character(len=*), intent(in) :: name
+    real(kind_phys),  intent(in) :: molar_mass   ! kg mol-1
+    integer,          intent(in) :: index_musica_species
+    integer,          intent(in) :: index_constituent_props
+    type(gas_species_t)          :: this
+    index_musica_species
+    this%name = name
+    this%molar_mass = molar_mass
+    this%index_musica_species = index_musica_species
+    this%index_constituent_props = index_constituent_props
+
+  end function species_t_constructor
+
+  !> Constructor for gas_species_t object
+  function species_profiled_t_constructor(name, unit, molar_mass, scale_height, &
+    index_musica_species, index_constituent_props) result( this )
+  
+    character(len=*), intent(in) :: name
+    character(len=*), intent(in) :: unit
+    real(kind_phys),  intent(in) :: molar_mass   ! kg mol-1
+    real(kind_phys),  intent(in) :: scale_height ! km
+    integer,          intent(in) :: index_musica_species
+    integer,          intent(in) :: index_constituent_props
+    type(gas_species_t)          :: this
+
+    this%name = name
+    this%unit = unit
+    this%molar_mass = molar_mass
+    this%scale_height = scale_height
+    this%index_musica_species = index_musica_species
+    this%index_constituent_props = index_constituent_props
+
+  end function species_profiled_t_constructor
+
+  subroutine cleanup_musica_species()
+
+    if (allocated( micm_species_set )) deallocate( micm_species_set )
+    if (allocated( micm_indices_constituent_props )) deallocate( micm_indices_constituent_props )
+    if (allocated( tuvx_species_set )) deallocate( tuvx_species_set )
+    if (allocated( tuvx_indices_constituent_props )) deallocate( tuvx_indices_constituent_props )
+    if (allocated( micm_molar_mass_array )) deallocate( micm_molar_mass_array )
+
+  end subroutine  cleanup_musica_species
+
+  subroutine register_musica_species(micm_species, tuvx_species)
+    type(musica_species_t), intent(in)  :: micm_species(:)
+    type(musica_species_t), intent(in)  :: tuvx_species(:)
+
+    number_of_micm_species = size(micm_species)
+    allocate( micm_species_set( number_of_micm_species ) )
+    micm_species_set = micm_species
+    number_of_micm_species = size(micm_species)
+    allocate( tuvx_elements_set( size(tuvx_species)) )
+    tuvx_species_set = tuvx_species
+
+  end subroutine register_musica_species
+
+  !> Retrieve the species indices from the constituents array and store them
+  subroutine find_musica_species_indices(constituent_props, musica_species_set, &
+      indices_constituent_props, errmsg, errcode)
+    use ccpp_constituent_prop_mod, only: ccpp_constituent_prop_ptr_t
+
+    type(ccpp_constituent_prop_ptr_t), intent(in)    :: constituent_props(:)
+    type(musica_species_t),            intent(inout) :: musica_species_set(:)
+    integer,                           intent(inout) :: indices_constituent_props(:)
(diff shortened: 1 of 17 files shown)


schemes/musica/musica_ccpp_species.F90
```diff
@@ -0,0 +1,322 @@
...
+
+    ! TODO(jiwon) Check molar mass is non zero as it becomes a denominator for unit converison
+    do i_elem = 1, size(micm_molar_mass_array)
+      if (micm_molar_mass_array(i_elem) <= 0) then
+        errcode = 1
+        errmsg = "[MUSICA Error] Molar mass must be greater than zero for " &
+                  // micm_species_set(i_elem)%name
+        return
+      end if
+    end do
```
boulderdaze, line 210: Hi @peverwhee, the last time we discussed this, we mentioned that the check could be done in the CCPP framework. Just wondering, do we have any updates on this?

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
```diff
@@ -0,0 +1,216 @@
...
+  character(len=*), parameter, public :: O2_LABEL = 'O2'
+  character(len=*), parameter, public :: O3_LABEL = 'O3'
+  character(len=*), parameter, public :: TUVX_GAS_SPECIES_UNITS = 'molecule cm-3'
+  real(kind_phys),  parameter, public :: SCALE_HEIGHT_DRY_AIR = 8.01_kind_phys ! km
+  real(kind_phys),  parameter, public :: SCALE_HEIGHT_O2 = 7.0_kind_phys       ! km
+  real(kind_phys),  parameter, public :: SCALE_HEIGHT_O3 = 7.0_kind_phys       ! km
+  !> Molar mass value of dry air is obtained from 'CAM-SIMA/src/utils/std_atm_profile.F90'
+  real(kind_phys),  parameter, public :: MOLAR_MASS_DRY_AIR = 0.0289644_kind_phys ! kg mol-1
+  real(kind_phys),  parameter, public :: MOLAR_MASS_O2 = 0.0319988_kind_phys      ! kg mol-1
+  real(kind_phys),  parameter, public :: MOLAR_MASS_O3 = 0.0479982_kind_phys      ! kg mol-1
```
boulderdaze, line 34: Hi @mattldawson, @nusbaume, after discussing with Kyle, I decided not to include the TUV-x gas species in the MICM configuration, as we are planning to move the parsing features to OpenAtmos, which will offer more flexibility. Currently, the MICM parser requires species and reaction JSON files, but TUV-x does not have reaction file.  I think it would be more efficient to use a separate configuration for TUV-x, as it requires additional component including cloud liquid water content and specific gas species such as O2, O3, and dry air. This module includes the constituents (or species) needed by TUV-x, with molar masses and scale heights hard-coded, as this file functions as a configuration. However, this setup may change in the future as the parser is moved to OpenAtmos.

schemes/musica/musica_ccpp_species.F90
peverwhee: hi @boulderdaze. Nope! I just made this issue - https://github.com/NCAR/ccpp-framework/issues/620 - to keep track of this and the other plan to check if it has been set.

Not sure when I'll be able to get to it, but at least we're tracking it!

boulderdaze: Sounds good, thank you!

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
nusbaume: Thanks for letting me know @boulderdaze!  I think as long as MUSICA can tell CAM-SIMA what the composition of the atmosphere is (ignoring physics or dynamics-specific constituents) then whatever method(s) you all choose is fine with me!

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
```diff
@@ -0,0 +1,216 @@
...
+    type(musica_species_t),                           intent(inout) :: micm_species(:)
+    type(musica_species_t),              allocatable, intent(out)   :: tuvx_species(:)
+    type(ccpp_constituent_properties_t), allocatable, intent(out)   :: constituent_props(:)
+    character(len=512),                               intent(out)   :: errmsg
+    integer,                                          intent(out)   :: errcode
+
+    ! local variables
+    integer                             :: num_new_species = 4
+    integer                             :: num_micm_species
+    type(ccpp_constituent_properties_t) :: temp_constituent_props(4)
```
jimmielin, line 56: Is this intended to be `temp_constituent_props(num_new_species)`? That might avoid issues with the # and the array size being out of sync in the future.

Could you also remind in the comments why a temporary constituent properties needs to be instantiated here? Thanks!

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
```diff
@@ -0,0 +1,216 @@
...
+      molar_mass = CLOUD_LIQUID_WATER_CONTENT_MOLAR_MASS, &
+      advected = .true., &
+      errcode = errcode, &
+      errmsg = errmsg )
+    if (errcode /= 0) return
+
+    ! Iterate through the MICM species to check if any TUV-x gas
+    ! species are included; if present, updates the scale height and profiled status.
+    do i_species = 1, num_micm_species
+      if (is_dry_air_registered .and. is_O2_registered .and. is_O3_registered) exit
```
jimmielin, line 85: Sorry for the basic question - could you elaborate on what's the difference between TUVx species and MICM species? Are the TUVx species the chemical tracers that are photolyzed (O3, NO2, HNO3, ...) or used to form the absorber columns? From having O2 and O3 I feel this is used for the absorber columns that feed into the photolysis?

I'm trying to understand if this is equivalent to the `... check for o2, o3 absorber columns` code in the old `mo_photo.F90` code in CAM-chem or it's initializing species that are photolyzed and passed into TUVx for photolysis rate computation. In the latter case the code seems to be very restrictive (since new photolysis reactions may be added in the future and it wouldn't be easy to hardcode these here for future updates)

jimmielin (changes requested):
Hi @boulderdaze, thanks for requesting my review. I have some minor comments and a few questions about the code.

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
boulderdaze: The constituents for TUVx include cloud liquid water content, along with O2, O3, and dry air, which are required to generate the profile and update their respective values. While some MICM configurations, such as the terminator, do not necessarily include O2 and O3, TUVx still requires these to define the constituents. I discussed this with Kyle, and we both agreed that it's better to use a separate configuration file rather than hard-coding these values. I  created a new[ issue](https://github.com/ESCOMP/atmospheric_physics/issues/189) to address this, as it doesn't seem straightforward, and there are several decisions to be made regarding how to generate profiles for O2, O3, and dry air, which will need to be handled differently from cloud liquid water content, NO and NO2. 
Yes, I agree this is a restrictive approach since this module functions as a configuration because we don't have a TUVx-specific parser. I'll address your comment in the mentioned issue.

schemes/musica/tuvx/musica_ccpp_tuvx_gas_species.F90
```diff
@@ -0,0 +1,147 @@
...
+    constituent_mol_per_cm_3(:) = constituents(:) * dry_air_density(:) / molar_mass / m_3_to_cm_3
+
+    num_vertical_levels = size(constituents)
+    interfaces(1) = constituent_mol_per_cm_3(num_vertical_levels)
+    interfaces(2:num_vertical_levels+1) = constituent_mol_per_cm_3(num_vertical_levels:1:-1)
+    interfaces(num_vertical_levels+2) = constituent_mol_per_cm_3(1)
+
+    if (tuvx_species_set(index_species)%name == O3_LABEL) then
+      densities(:) = height_deltas(:) * km_to_cm           &
+                  * ( interfaces(1:num_vertical_levels+1)  &
```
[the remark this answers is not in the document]
mattldawson: or maybe just indent the second line 2 spaces? My personal preference is to keep lines < ~80 characters

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
mattldawson: Sounds good to me too! We can discuss a longer-term solution when we move the parsing to OpenAtmos

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
boulderdaze: Yes, this module serves as a configuration to determine how many constituents need to be registered. O2 and O3 could be registered by MICM, and if so, they are not included in constituents_props. The temp_constituents_props definitely includes cloud liquid and only gas species that are not registered by MICM. By specifying the number of possible constituents, it avoids the need for memory allocation and deallocation at runtime. 
This can change once the TUVx configuration file is introduced. I will add comments explaining the purpose of temp_constituents_props. Could you review it?

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
mattldawson: These are the absorbing species needed to calculate the radiation field. Longer-term, I think we will want to make these configurable. But, as @boulderdaze commented above, we should probably do this once we finish moving to a combined configuration parser for all the MUSICA components.

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
```diff
@@ -0,0 +1,216 @@
...
+
+  public :: configure_tuvx_species, check_tuvx_species_initialization
+  
+  integer, protected, public :: index_cloud_liquid_water_content = MUSICA_INT_UNASSIGNED
+  integer, protected, public :: index_dry_air = MUSICA_INT_UNASSIGNED
+  integer, protected, public :: index_O2 = MUSICA_INT_UNASSIGNED
+  integer, protected, public :: index_O3 = MUSICA_INT_UNASSIGNED
+
+  ! Constants
+  ! Clould liquid water
```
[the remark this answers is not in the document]
boulderdaze: Fixed, thank you!

schemes/musica/tuvx/musica_ccpp_tuvx_gas_species.F90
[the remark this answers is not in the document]
boulderdaze: Yes, I’ve updated it. Could you let me know if it improves readability?
```
if (tuvx_species_set(index_species)%name == O3_LABEL) then
  densities(:) = height_deltas(:) * km_to_cm * 0.5_kind_phys &
      * ( interfaces(1:num_vertical_levels+1) + interfaces(2:num_vertical_levels+2) )
else
  densities(:) = height_deltas(:) * km_to_cm &
      * sqrt(interfaces(1:num_vertical_levels+1)) * sqrt(interfaces(2:num_vertical_levels+2))
end if
```

mattldawson (approved):
looks good!

schemes/musica/tuvx/musica_ccpp_tuvx_load_species.F90
jimmielin: Thanks @boulderdaze and @mattldawson!

schemes/musica/tuvx/musica_ccpp_tuvx_gas_species.F90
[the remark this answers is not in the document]
jimmielin: Looks good, thank you!

jimmielin (approved):
Thanks @boulderdaze for addressing my comments!

PR #185, opened 2024-12-24
don't set water species property for species that air_composition handles

peverwhee:
Originator(s): peverwhee

Removes water_species property from instantiate call for water vapor mixing ratio species (as air_composition will set those). We may run into this issue again at some point for the water vapor number concentration variables, but hopefully by then we'll have metadata properties for setting that!
can supersede https://github.com/ESCOMP/CAM-SIMA/pull/342

Describe any changes made to the namelist: n/a

M   test/test_schemes/initialize_constituents.F90
- remove water_species property from instantiate call

List any test failures: all expected tests pass

Is this a science-changing update? New physics package, algorithm change, tuning changes, etc? no

### DIFF
--- a/test/test_schemes/initialize_constituents.F90
+++ b/test/test_schemes/initialize_constituents.F90
@@ -119,7 +119,6 @@ subroutine initialize_constituents_register(constituents, errmsg, errcode)
              vertical_dim = 'vertical_layer_dimension', &
              min_value = 0.0_kind_phys,                 &
              advected = .true.,                         &
-             water_species = .true.,                    &
              mixing_ratio_type = 'wet',                 &
              errcode = errcode,                         &
              errmsg = errmsg)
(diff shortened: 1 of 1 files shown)


jimmielin (approved):
Thanks @peverwhee for fixing this!

test/test_schemes/initialize_constituents.F90
```diff
@@ -119,7 +119,6 @@ subroutine initialize_constituents_register(constituents, errmsg, errcode)
              vertical_dim = 'vertical_layer_dimension', &
              min_value = 0.0_kind_phys,                 &
              advected = .true.,                         &
-             water_species = .true.,                    &
```
nusbaume, line 122: Maybe worth adding a comment stating why we aren't setting the `water_species` property here?  Otherwise I feel like future us will forget and add it back in again.

nusbaume (approved):
Thanks @peverwhee!  Just have one comment request but otherwise everything looks good.

test/test_schemes/initialize_constituents.F90
jimmielin: Agreed @nusbaume I wasn't aware that `air_composition` could modify the water_species property after initialization so I thought it had to be set right here. A comment will be really helpful for when we forget about this one day :smile:

peverwhee: comment added. thanks guys!

PR #191, opened 2025-01-23
Add function for NO photolysis calculation

boulderdaze:


K20shores (approved):

schemes/musica/tuvx/musica_ccpp_tuvx_no_photolysis_rate.F90
```diff
@@ -21,79 +22,200 @@ module musica_ccpp_tuvx_no_photolysis_rate
...
     ! (see sphers routine description for details)
-    integer             :: number_of_crossed_layers(number_of_vertical_layers+1)
-    real(dk)            :: slant_path(0:number_of_vertical_layers+1,number_of_vertical_layers+1)
-    real(dk)            :: delta_z(number_of_vertical_layers+1) ! layer thickness (cm)
-    real(dk), parameter :: km2cm = 1.0e5_dk ! conversion from km to cm
-    real(dk)            :: jNO(number_of_vertical_layers) ! final photolysis rate
+    integer  :: number_of_crossed_layers(size(dry_air_density)+1)
+    real(dk) :: slant_path(0:size(dry_air_density)+1, size(dry_air_density)+1)
+    real(dk) :: delta_z(size(dry_air_density)+1) ! layer thickness (cm)
+    real(dk) :: jNO(size(dry_air_density)) ! final photolysis rate
```
boulderdaze, line 182: Hi @K20shores, @mattldawson, what do you think about renaming these variables such as?
 `delta_z` ->`layer_thickness`
 `jNO`-> `photolysis_rate_NO`

schemes/musica/tuvx/musica_ccpp_tuvx_no_photolysis_rate.F90
```diff
@@ -21,79 +22,200 @@ module musica_ccpp_tuvx_no_photolysis_rate
...
-    o3_dens(1) = o3_dens(2) * 7.0_dk / ( height_at_interfaces(1) - height_at_interfaces(2) )
+    call convert_mixing_ratio_to_molecule_cm3(constituents_O3, &
+          dry_air_density, MOLAR_MASS_O3, O3_densities(2:))
+    O3_densities(1) = O3_densities(2) * 7.0_dk / ( height_at_interfaces(1) - height_at_interfaces(2) )
 
-    call convert_mixing_ratio_to_molecule_cm3(constituents(:,NO_index), dry_air_density, molar_mass_NO, no_dens(2:))
-    no_dens(1) = no_dens(2) * 0.9_dk
+    call convert_mixing_ratio_to_molecule_cm3(constituents_NO_photolysis(:,index_NO), &
+          dry_air_density, MOLAR_MASS_NO, NO_densities(2:))
+    NO_densities(1) = NO_densities(2) * 0.9_dk
```
boulderdaze, line 201: Hi @mattldawson, @K20shores, I checked that in the CAM code, these (`7.0_dk`, `0.9_dk`, and so on) also appear to be magic numbers without explanation. Is there a way to figure out what they represent so we can define them as constexpr variables?

K20shores: I also don't know. I wonder if they are [scale heights](https://en.wikipedia.org/wiki/Scale_height), but I don't understand why they would be different. The original [paper](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/93JD02007) also didn't really indicate anything about this. We may have to ask Doug.

boulderdaze: Great, I asked him about it

schemes/musica/tuvx/musica_ccpp_tuvx_no_photolysis_rate.F90
K20shores: I like that

K20shores (approved):

PR #195, opened 2025-01-27
CCPPize Hack shallow convection

jimmielin:
Originator(s): @jimmielin

- CCPPize Hack shallow convection scheme (`cmfmca`) and related interstitials
- Move ZM namelist read to a separate `zm_conv_options` scheme so not all of ZM has to run while ZM namelist options are read for zm_conv_evap (needed for complete shallow convection)
- Freeze ChangeLog

Describe any changes made to the namelist:
- new hack shallow convection namelist options `cmfmca_namelist.xml`
- move ZM namelist options to `zm_conv_options_namelist.xml` 

CCPPize Hack shallow convection
A       schemes/hack_shallow/cmfmca.F90
A       schemes/hack_shallow/cmfmca.meta
A       schemes/hack_shallow/cmfmca_namelist.xml
A       test/test_suites/suite_convect_shallow_hack.xml
A       schemes/sima_diagnostics/convect_shallow_diagnostics.F90
A       schemes/sima_diagnostics/convect_shallow_diagnostics.meta

Interstitial schemes for zm_conv_evap
A       schemes/hack_shallow/set_general_conv_fluxes_to_shallow.F90
A       schemes/hack_shallow/set_general_conv_fluxes_to_shallow.meta
A       schemes/hack_shallow/set_shallow_conv_fluxes_to_general.F90
A       schemes/hack_shallow/set_shallow_conv_fluxes_to_general.meta

Interstitial schemes to combine shallow and deep convective fluxes
A       schemes/hack_shallow/convect_shallow_sum_to_deep.F90
A       schemes/hack_shallow/convect_shallow_sum_to_deep.meta

Move ZM namelist options to zm_conv_options scheme
A       schemes/zhang_mcfarlane/zm_conv_options.F90
A       schemes/zhang_mcfarlane/zm_conv_options.meta

Freeze ChangeLog
M       doc/ChangeLog

Move ZM namelist options to zm_conv_options scheme
R100    schemes/zhang_mcfarlane/zm_convr_namelist.xml   schemes/zhang_mcfarlane/zm_conv_options_namelist.xml
M       suites/suite_cam7.xml
M       test/test_suites/suite_zhang_mcfarlane.xml

List any test failures: 

Is this a science-changing update? New physics package, algorithm change, tuning changes, etc? 
New Hack shallow convection scheme


schemes/hack_shallow/cmfmca.F90
```diff
@@ -0,0 +1,978 @@
+! Copyright (C) 2024 National Science Foundation-National Center for Atmospheric Research
+! SPDX-License-Identifier: Apache-2.0
+!
+! Hack shallow convective scheme.
+!
+! Original Author: J. Hack
+! CCPPized: Haipeng Lin, October 2024
```
peverwhee, line 7: Can I request porting the full comment over from CAM, to preserve that history?

```suggestion
!-----------------------------------------------------------------------
!
! Purpose:
! Moist convective mass flux procedure:
!
! Method:
! If stratification is unstable to nonentraining parcel ascent,
! complete an adjustment making successive use of a simple cloud model
! consisting of three layers (sometimes referred to as a triplet)
!
! Code generalized to allow specification of parcel ("updraft")
! properties, as well as convective transport of an arbitrary
! number of passive constituents (see q array).  The code
! is written so the water vapor field is passed independently
! in the calling list from the block of other transported
! constituents, even though as currently designed, it is the
! first component in the constituents field.
!
! Author: J. Hack
!
! BAB: changed code to report tendencies in cmfdt and dq, instead of
! updating profiles. Cmfdq contains water only, made it a local variable
! made dq (all constituents) the argument.
!
! CCPPized: Haipeng Lin, October 2024
!-----------------------------------------------------------------------
```

schemes/hack_shallow/cmfmca.F90
```diff
@@ -0,0 +1,978 @@
...
+        endif
+      enddo
+    endif
+
+    ! in-module parameters (hardcoded)
+    dzmin = 0.0_kind_phys
+    betamn = 0.10_kind_phys
+
+    ! specify that relaxation timescale should be applied to column as opposed to triplets individually
+    rlxclm = .true.
```
peverwhee, line 116: can you just set these above at the module level, as you do `ssfac`, etc?

schemes/hack_shallow/cmfmca.F90
```diff
@@ -0,0 +1,978 @@
...
+    ! temporary for debug
+    use cam_logfile,               only: iulog
+
+    ! Input arguments
+    integer,         intent(in)     :: ncol               ! number of atmospheric columns
+    integer,         intent(in)     :: pver               ! number of vertical levels
+    integer,         intent(in)     :: pcnst              ! number of ccpp constituents
+    type(ccpp_constituent_prop_ptr_t), &
+                     intent(in)     :: const_props(:)     ! ccpp constituent properties pointer
+    integer,         intent(in)     :: nstep              ! current time step index
```
peverwhee, line 165: I don't think this is used - can remove from argument list and metadata.

schemes/hack_shallow/cmfmca.F90
```diff
@@ -0,0 +1,978 @@
...
+    errmsg, errflg &
+  )
+    ! framework dependency for const_props
+    use ccpp_constituent_prop_mod, only: ccpp_constituent_prop_ptr_t
+
+    ! to_be_ccppized
+    use wv_saturation,             only: qsat
+
+    ! temporary for debug
+    use cam_logfile,               only: iulog
```
peverwhee, line 157: Should this be an input argument to `cmfmca_run` ?

schemes/hack_shallow/cmfmca.meta
```diff
@@ -0,0 +1,281 @@
+[ccpp-table-properties]
+  name = cmfmca
+  type = scheme
```
peverwhee, line 3: i believe this should include the wv_saturation dependency

dependencies = ../../to_be_ccppized/wv_saturation.F90

schemes/hack_shallow/cmfmca.F90
```diff
@@ -0,0 +1,978 @@
...
+
+    ! Check constituents list and locate water vapor index
+    ! (not assumed to be 1)
+    const_check_loop: do m = 1, pcnst
+      call const_props(m)%standard_name(const_standard_name)
+      if (const_standard_name == 'water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water') then
+        const_wv_idx = m
+        exit const_check_loop
+      endif
+    enddo const_check_loop
```
peverwhee, line 329: I'd recommend using the `ccpp_const_get_idx` in `ccpp_const_utils` that you added. I hope that'll make it easier to find and replace with the framework version eventually...

schemes/hack_shallow/cmfmca_namelist.xml
```diff
@@ -0,0 +1,33 @@
...
+    <category>conv</category>
+    <group>hkconv_nl</group>
+    <standard_name>rain_water_autoconversion_coefficient_for_shallow_convection_tbd</standard_name>
+    <units>m-1</units>
+    <desc>
+      Rain water autoconversion coefficient for Hack shallow scheme.
+    </desc>
+    <values>
+      <!-- FIXME: This is a resolution-dependent value -->
+      <value>1.0e-4</value>
```
peverwhee, line 30: is this a "FIXME before merging this PR" or a "FIXME eventually when we have some capability? If it's the latter, I'd suggest a comment describing what we need to implement

schemes/sima_diagnostics/convect_shallow_diagnostics.F90
```diff
@@ -0,0 +1,307 @@
...
+    endif
+    if (const_cldice_idx > 0) then
+      call history_out_field('CMFDICE', dq(:,:,const_cldice_idx))
+    endif
+
+    ! CMFDQR
+    call history_out_field('CMFDQR', cmfdqr)
+
+    ! Calculate fractional occurrence of shallow convection
+    ! FIXME (hplin): this definition looks counter-intuitive but is replicated verbatim from convect_shallow.F90. To check.
```
peverwhee, line 192: did you check? :)

peverwhee (changes requested):
nothing massive - just a few requests! 

thanks @jimmielin woo! hack shallow!

schemes/sima_diagnostics/convect_shallow_diagnostics.F90
jimmielin: Thanks for the reminder, I did not! @adamrher, could you please help check if this is correct? This is from `convect_shallow.F90` - The definition of this just feels backwards to me, since it marks a column as having shallow convection when the shallow convection cloud mass flux is zero.
```fortran
    freqsh(:) = 0._kind_phys
    do i = 1, ncol
      if(maxval(cmfmc_sh(i,:)) <= 0._kind_phys) then
        freqsh(i) = 1._kind_phys
      endif
    enddo
    call history_out_field('FREQSH', freqsh)
```

schemes/hack_shallow/cmfmca.F90
jimmielin: Done, thanks! Added to above `cmfmca_run` (lightly edited to collapse in comment by BAB)

schemes/hack_shallow/cmfmca.F90
jimmielin: Thanks, done!

schemes/hack_shallow/cmfmca.F90
jimmielin: Done, thanks!

schemes/hack_shallow/cmfmca.F90
jimmielin: Yes, thanks! Updated

schemes/hack_shallow/cmfmca.F90
jimmielin: Thanks, updated!

schemes/hack_shallow/cmfmca_namelist.xml
jimmielin: Good question... it was more of a reminder that eventually the resolution-dependent options need to be ported, but from the [CAM-SIMA docs on namelist XML files](https://escomp.github.io/CAM-SIMA-docs/conversion/create-namelist-xml/) it looks like this work is implied: so I got rid of the comment.

> When populating the value fields, you don't need to port all of them at this time. You need to put in the value that is used by default by CAM, but the rest will be ported once all schemes have been ported.

schemes/hack_shallow/cmfmca.meta
jimmielin: Added, thank you!

schemes/sima_diagnostics/convect_shallow_diagnostics.F90
adamrher: Hi @jimmielin sorry I missed this git prompt. That is indeed strange -- a value of 1 should be assigned to the column when there is shallow convection. The comment above it "! Modification : I should check whether below computation of freqsh is correct." suggests we aren't the only one who thinks so (can you git blame this comment?).

I looked through hk_conv.F90 and verified that the array (cmfmc2, which I presume is cmfmc_sh in your code) cannot be negative. There are explicit limiters on the components that are summed and multiplied to give cmfmc2: beta and eta:

https://github.com/ESCOMP/CAM/blob/6eb3046284ef57385ca7612165465668c97ec1f5/src/physics/cam/hk_conv.F90#L697. 

And beta is limited using ```beta(i) = max(0.0_r8,beta(i))``` statements anytime it's modified. So I believe this is a bug in this particular diagnostic, and that the logic should be:
```
if (maxval(cmfmc_sh(i,:)) > 0._kind_phys) then
```

jimmielin: Thanks @adamrher! `git blame` shows that comment is before the Git era, unfortunately.

I've opened an issue in CAM: https://github.com/ESCOMP/CAM/issues/1254. I can work on a PR to fix this (probably a separate one from the CCPPization PR, so that the bit-for-bit property of the CCPPization PR can be preserved.)

cacraigucar: I went back to svn and a blame there yields:  
`mvr  4/29/2010 11:49:39 AM
updated the physics with code from the cam5 development branch; some changes to the dynamics, including the enabling of fourth-order divergence damping`
I don't remember who `mvr` is, but that person was the person who did the very first svn commit in CAM, so an experienced programmer.  Maybe Matt Rothstein (or maybe I'm making up names!)

swrneale: That was Matt's username. Oh well. Diagnostics can be tough to trouble shoot as the simulation may still be bfb.

jimmielin: Thanks @cacraigucar and @swrneale. It looks like FREQSH is not in the default field list for CAM4 so indeed if I change the definition of this diagnostic it will still show up as b4b in the regression tests. Anyway, I have fixed it on the CAM-SIMA end first (which has separate diagnostics code from current CAM) and can address it in ESCOMP/CAM separately.

adamrher: On second thought, it might not be worth the extra time fixing this diagnostic in ESCOMP/CAM, before you merge the PR to ESCOMP/CAM to call Hack from the atmospheric_physics version of convect_shallow.F90.

But I'm confused on one aspect. If we were running non-CCPP' cam5 physics in ESCOMP/CAM, which is also impacted by this bug, would it then use the convect_shallow in atmospheric_physics? Or are you separating just the cam4 parts out of convect_shallow for this Hack shallow task?

jimmielin: My apologies for the confusion! For this freqsh definition in particular, the code used for calculating it in CAM-SIMA is separate because it is in the "diagnostic" scheme, where the non-portable calls to the CAM history code reside. (`schemes/sima_diagnostics/convect_shallow_diagnostics.F90`, line ~191). The freqsh definition in ESCOMP/CAm will remain in convect_shallow.F90. So I would have to fix it in both atmospheric_physics and ESCOMP/CAM.

There is no direct equivalent to convect_shallow.F90 in atmos_phys -- `hk_conv.F90` (`subroutine cmfmca`) for Hack shallow is CCPPized (into `cmfmca.F90`), so the only part that is shared by CAM and CAM-SIMA is the `cmfmca_run` and `cmfmca_init` subroutines. 

There's a variety of other schemes in this PR, e.g., `schemes/hack_shallow/convect_shallow_sum_to_deep.F90` (combine shallow and deep convective fluxes to create "all" convective fluxes) and the diagnostics scheme (calls history_add_field and outfld) that do similar things as what `convect_shallow.F90` and `convect_diagnostics.F90` do, but in the SIMA way, so they're not used in ESCOMP/CAM (because history calls are different).

adamrher: OK I understand now. Thanks for straightening me out and I agree we'll need fix it in ESCOMP/CAM.

I actually really like this idea of dissolving the whole concept convect_shallow into portable interfaces and interstitials. It really helps clarify what parts of the code are used.

schemes/sima_diagnostics/convect_shallow_diagnostics.meta
```diff
@@ -0,0 +1,277 @@
...
+  units = kg kg-1 s-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = in
+[ qc_sh ]
+  standard_name = detrainment_of_water_due_to_shallow_convection
+  units = kg kg-1 s-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = in
```
nusbaume, line 265: We will also need to specify what `water` is for these variables here as well (i.e. `cloud_water_wrt_moist_air_and_condensed_water`).

test/test_suites/suite_convect_shallow_hack.xml
```diff
@@ -0,0 +1,63 @@
...
+    <!-- INITIALIZE DIAGNOSTICS (has init phase only) -->
+    <scheme>convect_shallow_diagnostics</scheme>
+
+    <!-- SHALLOW CONVECTION: HACK SCHEME -->
+    <scheme>check_energy_zero_fluxes</scheme>
+    <scheme>cmfmca</scheme>
+    <scheme>convect_shallow_diagnostics_after_shallow_scheme</scheme>
+    <scheme>apply_heating_rate</scheme>
+    <scheme>apply_constituent_tendencies</scheme>
+    <scheme>geopotential_temp</scheme>
```
nusbaume, line 37: We need to call `qneg` after `apply_constituent_tendencies`:

```suggestion
    <scheme>apply_constituent_tendencies</scheme>
    <scheme>qneg</scheme>
    <scheme>geopotential_temp</scheme>
```

test/test_suites/suite_convect_shallow_hack.xml
```diff
@@ -0,0 +1,63 @@
...
+    <scheme>cloud_fraction_fice</scheme>
+    <!-- prepare state for zm_conv_evap (rename shallow outputs to generic inputs) -->
+    <!-- also zero out quantities going in/out -->
+    <scheme>set_shallow_conv_fluxes_to_general</scheme>
+    <scheme>zm_conv_evap</scheme>
+    <scheme>set_general_conv_fluxes_to_shallow</scheme>
+    <scheme>convect_shallow_diagnostics_after_convective_evaporation</scheme>
+    <scheme>apply_heating_rate</scheme>
+    <scheme>apply_constituent_tendencies</scheme>
+    <scheme>geopotential_temp</scheme>
```
nusbaume, line 49: Need to call `qneg` here as well:

```suggestion
    <scheme>apply_constituent_tendencies</scheme>
    <scheme>qneg</scheme>
    <scheme>geopotential_temp</scheme>
```

test/test_suites/suite_convect_shallow_hack.xml
```diff
@@ -0,0 +1,63 @@
...
+    <scheme>check_energy_scaling</scheme>
+    <scheme>check_energy_chng</scheme>
+
+    <!-- DIAGNOSTIC QUANTITY UPDATES -->
+    <!-- eventually calculate shallow convective fraction diagnostic (FREQSH) -->
+    <!-- sum deep convective quantities with shallow including total convective rainout.
+         requires deep convective quantities to be actually available (i.e., rprddp)
+    <scheme>convect_shallow_sum_to_deep</scheme>
+    <scheme>convect_shallow_diagnostics_after_sum_to_deep</scheme>
+    -->
```
nusbaume, line 61: Remove commented-out code?  We likely won't be able to test these diagnostics in this particular test suite (as we only have the Hack outputs).

suites/suite_cam4.xml
```diff
@@ -0,0 +1,170 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<suite name="cam4" version="1.0">
+  <!-- 
+      CAM4 PHYSICS SUITE
+      Deep convection         ZM
+      Shallow convection      Hack
+      Macrophysics            RK    (not implemented)
+      Microphysics            RK    (not implemented)
+      Radiation               CAMRT (not implemented)
```
nusbaume, line 10: We will be using RRTMGP as the radiation package for all CAM suites, so I might update the comment here:

```suggestion
      Radiation               RRTMGP (not implemented)
```

suites/suite_cam4.xml
```diff
@@ -0,0 +1,170 @@
...
+    <scheme>qneg</scheme>
+    <scheme>geopotential_temp</scheme>
+    <scheme>cloud_fraction_fice</scheme>
+    <scheme>zm_conv_momtran</scheme>
+    <scheme>zm_momtran_tendency_diagnostics</scheme>
+    <scheme>apply_heating_rate</scheme>
+    <scheme>apply_tendency_of_eastward_wind</scheme>
+    <scheme>apply_tendency_of_northward_wind</scheme>
+    <scheme>qneg</scheme>
+    <scheme>geopotential_temp</scheme>
```
nusbaume, line 68: I don't believe we need to call `qneg` here (only after `apply_constituent_tendencies` is it needed).  This same change might be needed in `suite_cam7.xml` as well.

suites/suite_cam4.xml
```diff
@@ -0,0 +1,170 @@
...
+    <!-- INITIALIZE DIAGNOSTICS (has init phase only) -->
+    <scheme>convect_shallow_diagnostics</scheme>
+
+    <!-- SHALLOW CONVECTION: HACK SCHEME -->
+    <scheme>check_energy_zero_fluxes</scheme>
+    <scheme>cmfmca</scheme>
+    <scheme>convect_shallow_diagnostics_after_shallow_scheme</scheme>
+    <scheme>apply_heating_rate</scheme>
+    <scheme>apply_constituent_tendencies</scheme>
+    <scheme>geopotential_temp</scheme>
```
nusbaume, line 105: Need to call `qneg` here:

```suggestion
    <scheme>apply_constituent_tendencies</scheme>
    <scheme>qneg</scheme>
    <scheme>geopotential_temp</scheme>
```

suites/suite_cam4.xml
```diff
@@ -0,0 +1,170 @@
...
+    <scheme>cloud_fraction_fice</scheme>
+    <!-- prepare state for zm_conv_evap (rename shallow outputs to generic inputs) -->
+    <!-- also zero out quantities going in/out -->
+    <scheme>set_shallow_conv_fluxes_to_general</scheme>
+    <scheme>zm_conv_evap</scheme>
+    <scheme>set_general_conv_fluxes_to_shallow</scheme>
+    <scheme>convect_shallow_diagnostics_after_convective_evaporation</scheme>
+    <scheme>apply_heating_rate</scheme>
+    <scheme>apply_constituent_tendencies</scheme>
+    <scheme>geopotential_temp</scheme>
```
nusbaume, line 117: Need to call `qneg` here:

```suggestion
    <scheme>apply_constituent_tendencies</scheme>
    <scheme>qneg</scheme>
    <scheme>geopotential_temp</scheme>
```

suites/suite_cam4.xml
```diff
@@ -0,0 +1,170 @@
...
+
+    <!-- DIAGNOSTIC QUANTITY UPDATES -->
+    <!-- eventually calculate shallow convective fraction diagnostic (FREQSH) -->
```
nusbaume, line 120: I think I am confused by the "eventually" in this comment here.  Isn't `FREQSH` properly calculated in the diagnostics schemes now?

suites/suite_cam4.xml
```diff
@@ -0,0 +1,170 @@
...
+    -->
+
+    <!-- State diagnostics -->
+    <scheme>sima_state_diagnostics</scheme>
+
+    <!-- Find tropopause -->
+    <!-- Tropopause find has to make history output at the end of convective adjustment -->
+    <!-- so supersaturated cells are not written to the tape. -->
+    <scheme>tropopause_find</scheme>
+    <scheme>tropopause_diagnostics</scheme>
```
nusbaume, line 153: I believe this is called at the end of `physics_before_coupler` in CAM4.

suites/suite_cam4.xml
```diff
@@ -0,0 +1,170 @@
...
+  </group>
+
+  <group name="physics_after_coupler">
+
+    <!--
+      Vertical diffusion
+    -->
+
+    <!-- State diagnostics -->
+    <scheme>sima_state_diagnostics</scheme>
```
nusbaume, line 147: I believe this is called before radiation in the `physics_before_coupler` group in CAM4.

schemes/hack_shallow/cmfmca.meta
```diff
@@ -0,0 +1,294 @@
...
+  intent = in
+[ iulog ]
+  standard_name = log_output_unit
+  units = 1
+  type = integer
+  dimensions = ()
+  intent = in
+[ const_props ]
+  standard_name = ccpp_constituent_properties
+  units = None
```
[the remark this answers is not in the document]
peverwhee: popping in to say "yes please".

Also, some context: There was a time where all of the properties objects had to have `units = None` because the unit converter couldn't "convert" between None and none. But that is now fixed so lowercase all the way!

peverwhee (approved):
thanks @jimmielin !

schemes/hack_shallow/cmfmca.F90
```diff
@@ -0,0 +1,995 @@
+! Copyright (C) 2024 National Science Foundation-National Center for Atmospheric Research
+! SPDX-License-Identifier: Apache-2.0
+!
+! Hack shallow convective scheme.
+!
+! Original Author: J. Hack
+! CCPPized: Haipeng Lin, October 2024
+module cmfmca
```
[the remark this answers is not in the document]
nusbaume: Hi @jimmielin, I think `hack_convect_shallow` would be a great name for the scheme!

schemes/hack_shallow/cmfmca.meta
```diff
@@ -0,0 +1,294 @@
...
+  dimensions = (horizontal_loop_extent)
+  intent = out
+[ icwmr ]
+  standard_name = in_cloud_water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water_due_to_shallow_convection
+  units = kg kg-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ rliq_sh ]
+  standard_name = vertically_integrated_cloud_liquid_tendency_due_to_shallow_convection_to_be_applied_later_in_time_loop
```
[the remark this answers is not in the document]
nusbaume: Opening an issue is fine with me!

schemes/hack_shallow/cmfmca_namelist.xml
```diff
@@ -0,0 +1,32 @@
...
+    <type>real</type>
+    <category>conv</category>
+    <group>hkconv_nl</group>
+    <standard_name>rain_water_autoconversion_coefficient_for_shallow_convection</standard_name>
+    <units>m-1</units>
+    <desc>
+      Rain water autoconversion coefficient for Hack shallow scheme. Resolution dependent.
+    </desc>
+    <values>
+      <value>1.0e-4</value>
```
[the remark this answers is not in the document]
nusbaume: Thanks @jimmielin!  I think I would vote to leave the actual namelist definition file as it is now, and instead when we construct the CAM-SIMA regression test for it we can just make sure to set the namelist value to `1.0e-4` in the provided `user_nl_cam` file there.  That way the test will still work automatically but we won't have a special setting for the null dycore in the mainline source code.  Of course I am happy to hear any counterpoints you or others may have!

schemes/hack_shallow/convect_shallow_sum_to_deep.meta
```diff
@@ -0,0 +1,157 @@
...
+  dimensions = (horizontal_loop_extent, vertical_interface_dimension)
+  intent = in
+[ rprddp ]
+  standard_name = tendency_of_precipitation_wrt_moist_air_and_condensed_water_due_to_deep_convection_excluding_subcloud_evaporation
+  units = kg kg-1 s-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = in
+[ qc_deep ]
+  standard_name = detrainment_of_cloud_liquid_due_to_deep_convection
```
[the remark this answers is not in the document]
nusbaume: Hi @jimmielin, yep, the standard name for `dif` should also be updated to include the `wrt` suffix, good catch!  Of course feel free to just add that to an issue instead of fixing it in this PR if that is preferred.  Thanks!

schemes/hack_shallow/convect_shallow_sum_to_deep.meta
```diff
@@ -0,0 +1,157 @@
...
+  dimensions = (horizontal_loop_extent, vertical_interface_dimension)
+  intent = out
+[ rprdtot ]
+  standard_name = tendency_of_precipitation_wrt_moist_air_and_condensed_water_due_to_convection_excluding_subcloud_evaporation
+  units = kg kg-1 s-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ qc_total ]
+  standard_name = detrainment_of_water_due_to_all_convection
```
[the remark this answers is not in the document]
nusbaume: Thanks for the feedback!  Given that it sounds like we will have some CAM-induced confusion regardless I am fine to leave it is as `all` convection for now, although my guess is that eventually we'll need to convert it to just `convection` to follow the official standard name rules (but we can deal with that problem when we get there).

Going ahead and marking this comment as resolved.  Thanks!

schemes/sima_diagnostics/convect_shallow_diagnostics.meta
```diff
@@ -0,0 +1,277 @@
...
+[ cmfsl ]
+  standard_name = liquid_water_static_energy_flux_due_to_shallow_convection
+  units = W m-2
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_interface_dimension)
+  intent = in
+[ cmflq ]
+  standard_name = total_water_flux_due_to_shallow_convection
+  units = W m-2
```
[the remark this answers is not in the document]
nusbaume: Thanks for digging through the CAM code!  I agree that `W m-2` are the correct units here, so I will go ahead and mark this comment as resolved.

schemes/hack_shallow/cmfmca.meta
```diff
@@ -0,0 +1,294 @@
...
+  dimensions = ()
+  intent = in
+[ pref_edge ]
+  standard_name = reference_pressure_at_interface
+  units = Pa
+  type = real | kind = kind_phys
+  dimensions = (vertical_interface_dimension)
+  intent = in
+[ use_shfrc ]
+  standard_name = flag_for_cloud_area_fraction_use_shallow_convection_calculated_cloud_area_fraction
```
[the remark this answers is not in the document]
nusbaume: Looks like this standard name update still needs to be made?

schemes/hack_shallow/convect_shallow_sum_to_deep.F90
```diff
@@ -0,0 +1,124 @@
...
+      ! Input/output (total) arguments
+      ! In: deep only, Out: deep+shallow
+      real(kind_phys),   intent(inout) :: rliq_total(:)          ! Total convective reserved liquid [m s-1]
+
+      ! Output arguments
+      real(kind_phys),   intent(out)   :: cmfmc_total(:,:)       ! Total convective mass flux [kg m-2 s-1]
+      real(kind_phys),   intent(out)   :: rprdtot(:,:)           ! Total convective rainout Q tendency [kg kg-1 s-1]
+      real(kind_phys),   intent(out)   :: qc_total(:,:)          ! Total convection cloud water tendency [kg kg-1 s-1]
+      real(kind_phys),   intent(out)   :: cnt(:)                 ! Convective cloud top index [index]
+      real(kind_phys),   intent(out)   :: cnb(:)                 ! Convective cloud base index [index]
```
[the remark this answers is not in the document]
nusbaume: Looks like this question was missed in the previous review?

nusbaume (changes requested):
Thanks for resolving most of my questions/concerns @jimmielin!  I just had a few more questions, mostly related to missed comments from the previous review.

nusbaume (approved):
Thanks for implementing those final few suggestions @jimmielin!  I left un-resolved the two questions that are waiting on other people's input, but neither of them is required.  Thanks again!

test/test_suites/suite_zhang_mcfarlane.xml
```diff
@@ -25,7 +27,6 @@
        <scheme>apply_heating_rate</scheme>
        <scheme>apply_tendency_of_eastward_wind</scheme>
        <scheme>apply_tendency_of_northward_wind</scheme>
-       <scheme>qneg</scheme>
```
cacraigucar, line 28: Curious why qneg is being removed

cacraigucar (commented):
Concentrated my review on the Zhang-McFarlane impacts in this PR

cacraigucar (approved):

PR #204, opened 2025-02-14
Add kessler_test Test SDF for FPHYStest to split off from FKESSLER test

jimmielin:
Originator(s): jimmielin

The FPHYStest test for Kessler in SIMA right now uses the same `suite_kessler.xml` as FKESSLER, which means it includes `physics_after_coupler` and other `check_energy` schemes we do not want to include in the FPHYStest test which should only be testing `kessler_tend`.

This PR includes a new `suite_kessler_test.xml` that removes all the other schemes from the SDF and keeps only what we want to test, and consistent with how the snapshots are taken in CAM:

```fortran
      if (trim(cam_take_snapshot_before) == "kessler_tend") then
         call cam_snapshot_all_outfld(cam_snapshot_before_num, state, tend, cam_in, cam_out, pbuf)
      end if

      call kessler_tend(state, ptend, ztodt, pbuf)
      if ( (trim(cam_take_snapshot_after) == "kessler_tend") .and.            &
           (trim(cam_take_snapshot_before) == trim(cam_take_snapshot_after))) then
         call cam_snapshot_ptend_outfld(ptend, lchnk)
      end if
      call physics_update(state, ptend, ztodt, tend)

      if (trim(cam_take_snapshot_after) == "kessler_tend") then
         call cam_snapshot_all_outfld(cam_snapshot_after_num, state, tend, cam_in, cam_out, pbuf)
      end if
```

```
A       test/test_suites/suite_kessler_test.xml
  - new kessler_test suite
```

N/A

N/A

Yes for FPHYStest kessler test

BFB when using this suite

### DIFF
--- a/test/test_suites/suite_kessler_test.xml
+++ b/test/test_suites/suite_kessler_test.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<suite name="kessler_test" version="1.0">
+  <group name="physics_before_coupler">
+    <scheme>calc_exner</scheme>
+    <scheme>temp_to_potential_temp</scheme>
+    <scheme>calc_dry_air_ideal_gas_density</scheme>
+    <scheme>wet_to_dry_water_vapor</scheme>
+    <scheme>wet_to_dry_cloud_liquid_water</scheme>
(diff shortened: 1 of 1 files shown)


nusbaume (approved):
Thanks @jimmielin!

PR #228, opened 2025-04-03
Fix broken github action tests

cacraigucar:
Originator(s): haipeng, mwaxmonsky, cacraig

Description (include issue title and the keyword ['closes', 'fixes', 'resolves'] and issue number): N/A

M       .github/workflows/test.yaml
M       .github/workflows/unit-tests.yaml
             - Changed ubuntu to a fixed version instead of "latest" and then reverted it back to ubuntu-latest, so no
                effective changes 
             
M       test/docker/Dockerfile.musica
M       test/docker/Dockerfile.musica.no_install
M       test/musica/CMakeLists.txt
            - Changes necessary to support the changed github repository name for CCPP Standard Names
           
Is this an answer-changing PR? If so, is it a new physics package, algorithm change, tuning change, etc?  No

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,7 +4,7 @@ on: [pull_request,workflow_dispatch]
 
 jobs:
   test_musica_api:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
       with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,7 +4,7 @@ on: [pull_request,workflow_dispatch]
 
 jobs:
   test_musica_api:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4
       with:
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   gcc-toolchain:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout atmospheric_physics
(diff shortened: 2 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -4,7 +4,7 @@ on: [pull_request,workflow_dispatch]
 
 jobs:
   test_musica_api:
-    runs-on: ubuntu-latest
```
nusbaume, line 7: I would leave this workflow as `ubuntu-latest`, as it doesn't depend on the pFUnit library.

nusbaume (changes requested):
Looks good!  Just had one change request.

.github/workflows/test.yaml
cacraigucar: @mwaxmonsky requested I change both - let's see what he has to say

mwaxmonsky: Ah okay, apologies for my uninformed advice!

After taking a closer look, if this works using `ubuntu-latest` (which it looks like it does from the commit prior to this change: https://github.com/ESCOMP/atmospheric_physics/actions/runs/14246373168/job/39928308391), then I would agree with sticking with `ubuntu-latest` here as well (and maybe hard code it to `ubuntu-24.04 just so we don't get surprised by these moving forward?) and if there are issues, default back to `ubuntu-22.04`.

Though that would break the consistency among the CI (e.g., we will have some workflows working on latest and greatest image versions, and other workflows on previous versioned images).  It just depends on what we want our CI philosophy to be (e.g., allow differing version support or prefer consistent versioning throughout), but that's probably a discussion to be had outside this PR.

cacraigucar: Neither of these needed to be changed to a hardcoded, I believe.  I am going to put both back to ubunt-latest and if the tests pass, I will propose we move this to a new issue?

nusbaume (approved):

PR #229, opened 2025-04-03
Handle multiple calls to qneg

peverwhee:
Originator(s): peverwhee

**Handle multiple calls to all qneg phases. Since it's not yet clear how the framework will handle multiple instances of the same scheme in an SDF, this PR protects against multiple calls to the qneg summary printer.**

List all namelist files that were added or changed: n/a

**M   schemes/utilities/qneg.F90**
- **Add module-level variables to handle multiple calls to `qneg_timestep_final` and `qneg_final`**

Is this an answer-changing PR? If so, is it a new physics package, algorithm change, tuning change, etc? No.


nusbaume (approved):
Thanks @peverwhee!

schemes/utilities/qneg.F90
```diff
@@ -301,11 +306,18 @@ subroutine qneg_timestep_final(mpi_communicator, rootprocid, isrootproc,   &
       errcode = 0
       errmsg = ''
 
+      if (qneg_timestep_finalized) then
+         ! Only finalize the timestep once even if qneg_timestep_final is called multiple times
+         return
+      end if
+
```
cacraigucar, line 313: What should happen if qneg is finalized more than once?  Will the printed summary just be from the first place it is called?

peverwhee: Yes, which should be OK because all of the qneg calculations that will be printed out in the final() call will have been done during the run phase.

cacraigucar (approved):

PR #240, opened 2025-04-23
Initial file reader object attempt

### LINKED ISSUE #241: Need to provide example of a host-agnostic File I/O interface
Several physics schemes that are or will become CCPP-compliant require various forms of non-decomposed File I/O, often to set input data or parameters that are specific to it or related physics schemes.

The CCPP standards allow for file I/O during init and finalize phases.  However, this currently requires the physics schemes to mange the I/O itself directly, for example making direct Fortran NetCDF calls in the scheme's init phase, which can be a burden for physics developers.  Additionally,  many host models may have their own internal infrastructure or libraries to handle  file I/O, although those libraries tend to be host model-specific (e.g. the PIO library in CAM-SIMA).

One way to solve these issues is to define a host-model agnostic file I/O interface that physics schemes can use to manage their file data.  Then the host model can connect that generic interface to an actual implementation in the host model that can use model-specific libraries.   This allows for physics schemes to be portable (a CCPP goal) while also allowing for the use of host-specific I/O libraries.

A first attempt will be made here in the ESCOMP/atmospheric_physics and ESCOMP/CAM-SIMA repos.  Once that is done an issue will be open in the main NCAR/ccpp-framework repo  pointing to this example with additional discussions and possible modifications, with the hope that eventually the host-agnostic interface routines can live in the CCPP-framework repo itself.

nusbaume:
Originator(s): @mwaxmonsky @nusbaume 

This PR contains a File I/O class as designed by myself and (mostly) @mwaxmonsky.  This current version should be able
to read in any integer, real, or character-type variable from a NetCDF file, up to a five-dimensional array.  Currently
the entire variable will be read, but a future PR will also allow for variable subsetting, particularly for the record dimension
(as that is required for multiple CAM-SIMA use cases).

A few notes on the design:

1.  The major goal of this work is to create a File I/O interface that is host model independent.  We believe
    we have accomplished that but if you see anything that appears SIMA-specific please let us know!

2.  The class is designed to allocate a variable's memory on the host model side.  This was done with the idea
    that a host model might have a more sophisticated way of managing memory then just putting it on the heap (e.g. a memory pool).

3.  We were torn on how many capabilities the FileIO class should have.  We decided to keep the functionality low to reduce
    complexity, but if there are strong feelings otherwise please let us know!  In particular we are leaning towards making
    the necessary regridding/interpolation steps a separate class, so that we don't have too much inter-mingling of functionality
    in a single location.

Anyways, any thoughts, comments, suggestions, or review requests would be greatly appreciated!  The corresponding CAM-SIMA PR is ESCOMP/CAM-SIMA#385.  Thanks!

Fixes #241

A       test/test_schemes/file_io_test_namelist.xml
 - Namelist file for file I/O test scheme which can be used to specify the test NetCDF file(s).

List all files eliminated and why: N/A (no files removed)

A       phys_utils/ccpp_io_reader.F90
 - Initial File I/O reader type declarations
 
A       test/test_schemes/file_io_test.F90
A       test/test_schemes/file_io_test.meta
A       test/test_suites/suite_file_io_test.xml
 - File I/O test scheme and suite for use in CAM-SIMA testing (will be removed once a more complete/automated testing system has been developed).

List all existing files that have been modified, and describe the changes:  N/A (no files were modified).

List all automated tests that failed, as well as an explanation for why they weren't fixed: All tests pass.


test/test_schemes/file_io_test.F90
K20shores, line 1: `phys_utils/ccpp_io_reader.F90` contains an interface class, but it's used directly here. I don't understand how. is there a missing file, maybe, or is this yet another gap of ignorance in my fortran knowledge?

jimmielin (approved):
Thanks @mwaxmonsky @nusbaume! This looks great! I just had one question for test code in this PR and I have another comment in the CAM-SIMA PR.

test/test_schemes/file_io_test.F90
mwaxmonsky: Just to follow up from our discussion, the implementation exists in the host via a fortran `submodule`. The `ccpp_io_reader.F90` just contains the interface and the `create_io_reader_t()` call is overridden in the host via those submodules that are built/linked in specifically for the chosen backend.  The build in CAM-SIMA is hard coded to only support pio at the moment but we will implement the logic to build against netcdf and include a raw netcdf library implementation next.

test/test_schemes/file_io_test.F90
```diff
@@ -0,0 +1,116 @@
...
+     integer :: absorber
+
+     reader = create_io_reader_t()
+
+     !Initialize output variables:
+     errcode = 0
+     errmsg  = ''
+
+     ! Open file
+     call reader%open_netcdf_file(file_path, errcode, errmsg)
```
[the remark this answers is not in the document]
mwaxmonsky: So I need to test this but _in theory_, you might be able to reuse the reader object so long as ordered `open`/`close` calls are made, you could possibly reuse the same reader object but my initial thought process was one reader object per file just to make it easier on the accounting end.

At the current moment, this will work fine for a one time init call but might be problematic on a per timestep basis especially if there are a large quantity of files needed to be read in which case a single reusable io object might be more performant.

There is also the question of parallel io operations and this design definitely doesn't take that requirement into account, that is if we need to read in large numbers of files each time step.

But ultimately I'll have defer to @nusbaume or a discussion with the rest of the stakeholders as to what the ultimate requirements are for this library.

boulderdaze (approved):
The code looks good to me! The review thread helped me understand it better.

gold2718:
Is there an issue somewhere with a description of what this tools is supposed to do and where it can be used in CCPP physics schemes? For instance, is this for reading decomosed or non-decomposed data? 

In case it is useful, here are some snippets from the original CCPP requirements (is CAM-SIMA following these requirements? something else?):
- `C11`: At initial, restart, or finalize time parameterizations may read non-decomposed data (such as look-up
tables) that do not have scope outside of the physics scheme.
- `H3`: The host application cap (or the upstream calling application) shall perform all required I/O, except for a few notable cases (see C 11 ).

mwaxmonsky:
> Is there an issue somewhere with a description of what this tools is supposed to do and where it can be used in CCPP physics schemes? For instance, is this for reading decomosed or non-decomposed data?
> 
> In case it is useful, here are some snippets from the original CCPP requirements (is CAM-SIMA following these requirements? something else?):
> 
> * `C11`: At initial, restart, or finalize time parameterizations may read non-decomposed data (such as look-up
>   tables) that do not have scope outside of the physics scheme.
> * `H3`: The host application cap (or the upstream calling application) shall perform all required I/O, except for a few notable cases (see C 11 ).
> * `H4`: The host application cap (or the upstream calling application) shall handle all required run-time distributed memory processing for decomposed arrays, if any is required. Upon entry into each parameterization, each input field in the argument list is assumed to be the correct value to use

The only requirements I was going off of was the need for physics code to be able to read in file data at init and timestep init time.  Beyond that, any documented requirements are news to me and I would have to default to @mattldawson @nusbaume and @peverwhee for the details there.

Do you have a link to the requirements document that I could look at?

phys_utils/ccpp_io_reader.F90
```diff
@@ -0,0 +1,70 @@
...
+    type, abstract :: ccpp_io_reader_t
+    contains
+        procedure(open_netcdf_file),    deferred :: open_netcdf_file
+        procedure(close_netcdf_file),   deferred :: close_netcdf_file
+        procedure(get_netcdf_var_int),  deferred :: get_netcdf_var_int
+        procedure(get_netcdf_var_real), deferred :: get_netcdf_var_real
+        procedure(get_netcdf_var_char), deferred :: get_netcdf_var_char
+
+        generic :: get_netcdf_var => get_netcdf_var_int, get_netcdf_var_real, get_netcdf_var_char
+    end type ccpp_io_reader_t
```
mattldawson, line 17: This entire comment is very minor, so feel free to ignore the whole thing. But, and I may be misunderstanding this, is the idea that this class is specific to reading NetCDF files? If so, would it make sense to just name the class something like `netcdf_reader_t`, and remove `netcdf` from all the functions?

I understand having the prefix `ccpp_` for the module name to prevent module name conflicts, but is it really needed for the class name? It doesn't seem to add much information for a user (unless I'm missing something).

Also, is the longterm goal to have output capabilities as part of this class? If so, maybe `file_io_t` or `netcdf_io_t`? Otherwise, maybe `file_reader_t` or `netcdf_reader_t`? Having both IO and reader seems inconsistent to me.

phys_utils/ccpp_io_reader.F90
```diff
@@ -0,0 +1,70 @@
...
+
+    interface
+        module function create_io_reader_t() result(r)
+            class(ccpp_io_reader_t), allocatable :: r
+        end function create_io_reader_t
+
+        subroutine open_netcdf_file(this, file_path, errcode, errmsg)
```
mattldawson, line 24: I think I've seen in other parts of the code the `errmsg` come before `errcode`, but I'm fine with this ordering, just thought I'd mention it.

phys_utils/ccpp_io_reader.F90
```diff
@@ -0,0 +1,70 @@
...
+            class(ccpp_io_reader_t), intent(inout)  :: this
+            integer,                 intent(out) :: errcode !Error code
+            character(len=*),        intent(out) :: errmsg
+        end subroutine close_netcdf_file
+
+        subroutine get_netcdf_var_int( this, varname, var, errcode, errmsg)
+            import ccpp_io_reader_t
+            class(ccpp_io_reader_t), intent(in)  :: this
+            character(len=*),        intent(in)  :: varname
+            integer, pointer,        intent(out) :: var(..) !Character variable that file data will be read to.
```
mattldawson, line 44:
```suggestion
            integer, pointer,        intent(out) :: var(..) !Integer variable that file data will be read to.
```

phys_utils/ccpp_io_reader.F90
```diff
@@ -0,0 +1,70 @@
...
+            integer,                 intent(out) :: errcode !Error code
+            character(len=*),        intent(out) :: errmsg  !Error message
+        end subroutine get_netcdf_var_int
+
+        subroutine get_netcdf_var_real( this, varname, var, errcode, errmsg)
+            use ccpp_kinds, only: kind_phys
+            import ccpp_io_reader_t
+            class(ccpp_io_reader_t),       intent(in)  :: this
+            character(len=*),              intent(in)  :: varname
+            real(kind_phys), pointer,      intent(out) :: var(..) !Character variable that file data will be read to.
```
mattldawson, line 54:
```suggestion
            real(kind_phys), pointer,      intent(out) :: var(..) !Floating-point variable that file data will be read to.
```

mattldawson (approved):
Nice! I just a had a couple very minor comments.

Following up on @jimmielin 's comments, if parallel use of multiple instances of the file reader class and/or repeated use of a single instance for multiple files are wanted, I would recommend adding those use cases to the tests.

Other than that, it looks great! I might have more questions once I see the PIO-backed extending class.

gold2718:
> The only requirements I was going off of was the need for physics code to be able to read in file data at init and timestep init time

@mwaxmonsky, thanks for the quick response. Is the file data limited to non-decomposed data as in the sample test code? That certainly does not seem to go against any of the [requirements](https://dtcenter.org/sites/default/files/community-code/ccpp-requirements-historical.pdf) I copied.

K20shores (approved):

test/test_schemes/file_io_test.F90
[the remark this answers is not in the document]
mwaxmonsky: @jimmielin Based on our conversation, it sounds like we're matching performance with HEMCO requiring a single file per reader object that is created/destructed for each use.

If these become runtime bottlenecks, we can adjust our interface then and take into account the multiple file lifetimes aspect as well.

phys_utils/ccpp_io_reader.F90
mwaxmonsky: Updated!

nusbaume:
@gold2718 Sorry about that!  I created issue #241 to explain the original reasoning behind this PR.  Ultimately once this is all setup here we'll be opening an issue in the actual CCPP framework for additional discussion (but wanted a fully working example first).

phys_utils/ccpp_io_reader.F90
mwaxmonsky: Updated!

Changed the reader name to be `abstract_netcdf_reader_t` to specify that it is abstract and specifically meant for netcdf files.  Removed the `netcdf` test from the API functions.

Left it as `reader` explicitly to imply that it is only for reading and nothing else as currently we don't want physics code changing files in the middle of physics but that functionality may be desired for different reasons, but at that point, that might be a different object that exists purely in the host.

phys_utils/ccpp_io_reader.F90
mwaxmonsky: Good point!  Updated!

nusbaume:
Hi @kuanchihwang, just checking if you were planning to review this PR as well?  Certainly no rush, I just didn't know if you got the notification or not (and please note that your review isn't required, I just wanted to give you the chance as it might impact MMM physics schemes).  Thanks!

kuanchihwang (approved):
@nusbaume Looks great to me! A few MMM physics schemes indeed need to read data during initialization. As far as I know, those data are not time-stepped, so the use case should be well-covered by this PR.

Apologize that I got caught up in other administrative tasks.

PR #248, opened 2025-05-15
Break up file i/o object into explicit interfaces

peverwhee:
Originator(s): peverwhee

Unfortunately, due to inconsistent compiler support, this PR removes the fancy assumed rank parts of the file i/o object and replaces them with an interface for each dimension/type.

List all namelist files that were added or changed: n/a

M phys_utils/ccpp_io_reader.F90
- Break into dimension-specific interfaces

List all automated tests that failed, as well as an explanation for why they weren't fixed: n/a

Is this an answer-changing PR? If so, is it a new physics package, algorithm change, tuning change, etc? No

### DIFF
--- a/phys_utils/ccpp_io_reader.F90
+++ b/phys_utils/ccpp_io_reader.F90
@@ -9,11 +9,28 @@ module ccpp_io_reader
     contains
         procedure(open_file),    deferred :: open_file
         procedure(close_file),   deferred :: close_file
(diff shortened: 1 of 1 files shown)


nusbaume (approved):
Thanks @peverwhee!  This looks good to me (even if I still haven't fully emotionally recovered yet).

mwaxmonsky (approved):
Thanks @peverwhee!!

PR #252, opened 2025-06-13
Update MUSICA suite

mattldawson:
Originator(s): Matt Dawson

- Updates the MUSICA suite file to include diagnostics so we can include chemical species in the output files.
- Updates the MICM wrapper to output warnings to the log file when the solver fails.
- Updates to the latest version of the MUSICA library

none

none

none

none

It should only affect answers for chemistry


nusbaume (approved):
Looks good to me too.  Thanks @mattldawson!

PR #264, opened 2025-07-24
CCPP framework only allows standard names as dimension indices

peverwhee:
Originator(s): peverwhee

With the new CCPP-framework updates, we can no longer use integer dimension indices (like 1:horizontal_loop_extent).

This PR removes a couple of those.

M   schemes/sima_diagnostics/zm_diagnostics.meta
M   schemes/sima_diagnostics/zm_evap_tendency_diagnostics.meta
- remove "1:" in metadata dimensions

Is this an answer-changing PR? If so, is it a new physics package, algorithm change, tuning change, etc? No.


cacraigucar (approved):
I would suggest changing the title to be "CCPP framework no longer allows indices containing dimension ranges".  My first thought was how will we get by without integer dimensions?

peverwhee:
@cacraigucar how about "numerical"? The problem isn't ranges it's the use of actual numbers

cacraigucar:
> The Fortran term is `literal`. Another way to phrase it is: CCPP framework only allows standard names as dimension indices

I like this suggestion - I now understand the purpose of this PR from its title

nusbaume (approved):
Thanks @peverwhee!

PR #296, opened 2025-09-08
Set tendencies to zero if rayleigh friction if turned off

fvitt:
modified:   schemes/rayleigh_friction/rayleigh_friction.F90

Originator(s):

List all `development` PR numbers included in this PR and the title of each:


nusbaume:
I believe this PR needs to go into the development branch, at which point we can make a development to main PR for CAM (which I can do).  @fvitt Are you ok with that?

Issue #118, opened 2024-09-13
CCPP-ize PUMAS


nusbaume:
This issue is here to help track the CCPP-ization of the current CAM interface to the PUMAS microphysics package, with the eventual goal of having a version of PUMAS runnable in CAM-SIMA.

Also notifying @Katetc and @andrewgettelman that I'll start work on this soon, and so may have some questions for them at some point.

nusbaume:
The following is a list of needed tasks to assist with tracking progress on PUMAS CCPP-ization:

Tasks needed explicitly in the [ESCOMP/PUMAS](https://github.com/ESCOMP/PUMAS) repo:

- [ ] Remove all uses of CESM-specific source code from internal PUMAS routines (except the `qsat_X` routines, which will need to be a dependency).
- [ ] Create a CCPP cap for the main PUMAS routine (and ensure it works in CAM).
- [ ] Create Metadata files for the PUMAS CCPP cap and the "proc_rates_type" DDT.
- [ ] Create Github Action which validates the metadata files whenever a PR is made.
- [ ] Add PUMAS submodule to NCAR/ccpp-physics repo, along with needed dependencies.

Tasks needed for a complete PUMAS interstitial scheme that lives in [NCAR/ccpp-physics](https://github.com/NCAR/ccpp-physics):

- [ ] Register all constituents needed by PUMAS.
- [ ] Subset variables into expected microphysics dimensions (e.g. subcolumns, when enabled).
- [ ] Convert microphysics-dimensioned variables back into regular-dimensioned variables.
- [ ] Apply any associated tendencies to their states,  possibly in several subcycling steps.
- [ ] Calculate (and apply if necessary) the "final", post-subcycle microphysics tendencies.
- [ ] Calculate size distribution parameters and Liquid Water Path needed for diagnostics and radiation (i.e. the `size_dist_param_X` subroutine calls in CAM)
- [ ] Calculate remaining diagnostic variables, such as precipitation efficiency.
- [ ] Output diagnostics using CAM-SIMA-specific method(non-portable)

Tasks needed to fully-implement a CCPP-ized PUMAS in [ESCOMP/CAM-SIMA](https://github.com/ESCOMP/CAM-SIMA) :

- [ ] Create a CAM-SIMA-specific `pumas_namelist.xml` file to support PUMAS namelist variables in CAM-SIMA.
[...]

PR #357, opened 2026-01-16
atmos_phys0_27_000: PUMAS round3

cacraigucar:
Tag name (The PR title should also include the tag name): atmos_phys0_27_000
Originator(s): cacraigucar, nusbaume, billsacks, peverwhee, I-Kuan Hu, jimmielin

  - PUMAS code now has metadata and goes through the CAM-SIMA compilation step.  [Note PUMAS diagnostics will come in a later stage]  - This step has been named "round3".
  - This commit contains contributions from participants at the PUMAS hackathon 
  - Partially addresses #118 

A       schemes/pumas/pumas_pre_main_namelist.xml
            - pumas CAM-SIMA namelist generator 

A       schemes/pumas/micro_pumas_ccpp_dimensions_post.F90
A       schemes/pumas/micro_pumas_ccpp_dimensions_post.meta
A       schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90
A       schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
            - code which subsets and converts to pumas_r8 and back
            
A       schemes/pumas/pumas_pre_main.F90
A       schemes/pumas/pumas_pre_main.meta
A       schemes/pumas/pumas_post_main.F90
A       schemes/pumas/pumas_post_main.meta
            - placeholder for interstitial schemes
            
A       schemes/sima_diagnostics/pumas_diagnostics.F90
A       schemes/sima_diagnostics/pumas_diagnostics_calling.F90
A       schemes/sima_diagnostics/pumas_diagnostics.F90-keep_for_run
            - These are being included even though they are only partially complete - NO REVIEW IS BEING DONE NOW

A       test/test_suites/suite_pumas.xml
            - test suite for PUMAS

A       schemes/microp_aero/ccpp/hetfrz_classnuc_stub.F90
A       schemes/microp_aero/ccpp/hetfrz_classnuc_stub.meta
A       schemes/microp_aero/ccpp/hetfrz_classnuc_stub_namelist.xml
            - stub for hetfrz_classnuc to provide the namelist option do_hetfrz_classnuc used by PUMAS core (micro_pumas_ccpp)

A       schemes/pumas/micro_pumas_ccpp_optics_limiter.F90
A       schemes/pumas/micro_pumas_ccpp_optics_limiter.meta
            - optics property limiter for strat_cldfrc<cldfrc_min copied from micro_pumas_cam

A       schemes/utilities/clamp_number_concentrations.F90
A       schemes/utilities/clamp_number_concentrations.meta
            - clamp number concentrations to 1e-12, 1e10 range the same as CAM physics types

A       test/test_schemes/clubb_stub.F90
A       test/test_schemes/clubb_stub.meta
             - stub for CLUBB

A       schemes/pumas/pumas_pre_main_namelist.xml
             - namelist for PUMAS

M       .gitmodules
M       doc/NamesNotInDictionary.txt
M       schemes/pumas/pumas

M       suites/suite_cam5.xml
M       suites/suite_cam7.xml
             - Add PUMAS to CAM5 and CAM7 suites

M       schemes/rasch_kristjansson/rk_stratiform.meta
M       schemes/rrtmgp/rrtmgp_lw_cloud_optics.meta
M       schemes/rrtmgp/rrtmgp_sw_cloud_optics.meta
M       schemes/sima_diagnostics/rk_stratiform_diagnostics.F90
M       schemes/sima_diagnostics/rk_stratiform_diagnostics.meta
M       test/test_schemes/initialize_constituents.F90
              - updated standard names

(

    - new capability in CAM-SIMA


jimmielin:
Please note that since `compute_cloud_fraction_two_moment` has been merged, `rhmini` in `micro_pumas_ccpp` can now use its value via the namelist and the hardcoded value in registry.xml for CAM-SIMA can also be removed. Please see https://github.com/nusbaume/PUMAS/pull/1

nusbaume:
## Automated code review (PR #357 — PUMAS round3)

This is an AI-assisted review (`claude-opus:4.8`) against the rules in `doc/Code-review.md`, `doc/Code-style.md`, and `AGENTS.md`. Per the PR description, `pumas_diagnostics.F90` and `pumas_diagnostics_calling.F90` are explicitly *not* being reviewed at this stage, so I've excluded them (I did note a couple of things about them at the bottom for later). Auto-generated `doc/NamesNotInDictionary.txt` and the submodule pointer/`.gitmodules` bumps were not reviewed.

### High priority

1. **Missing `implicit none` — `schemes/pumas/micro_pumas_ccpp_dimensions_post.F90`** (Code-style MUST). The module preamble goes straight from the `use` statements to `public ::` with no `implicit none`. Every module must have `implicit none` in the preamble.

2. **`errmsg` must be `character(len=*)`, not `character(len=512)` (Code-review rule 11).** The rule explicitly covers CCPP-framework variables such as `errmsg`. This affects:
   - `schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90` (both subroutines)
   - `schemes/pumas/micro_pumas_ccpp_dimensions_post.F90`
   - `schemes/pumas/micro_pumas_ccpp_optics_limiter.F90`
   - `schemes/pumas/pumas_pre_main.F90` (both subroutines)
   - `schemes/pumas/pumas_post_main.F90`
   - `schemes/utilities/clamp_number_concentrations.F90` (both subroutines)
   - `test/test_schemes/clubb_stub.F90`

   For reference, `schemes/microp_aero/ccpp/hetfrz_classnuc_stub.F90` (added in this same PR) already does this correctly with `character(len=*)`.

3. **Module-level `save` — `schemes/utilities/clamp_number_concentrations.F90:13` (Code-review rule 1).** Module variables are implicitly `save`d, so the explicit module-level `save` is redundant and should be removed.

### Medium priority

4. **`endif` → `end if` (Code-review rule 4)** — `schemes/pumas/pumas_pre_main.F90:46` (the `if (do_clubb_sgs)` block).

5. **Commented-out code (Code-review rule 13) — `schemes/utilities/clamp_number_concentrations.F90`.** The `_run` subroutine ends with a large block of commented-out alternative implementation (the `q_projected = ...` formulation, ~30 lines). The explanatory rationale is valuable, but per rule 13 the dead code itself should be removed (keep the explanation as prose if you want the reasoning preserved). Note this also leaves the declared local `q_projected` unused in the active code.

6. **Namelist XML missing the standard comment block (Code-review rule 6).** Neither new namelist file has the standard `<!-- Each namelist variable is defined in an <entry> element... -->` block that follows `<entry_id_pg version="2.0">` in the other namelist XMLs (e.g. `dadadj_namelist.xml`):
   - `schemes/microp_aero/ccpp/hetfrz_classnuc_stub_namelist.xml`
   - `schemes/pumas/pumas_pre_main_namelist.xml`

7. **Missing default `private` + explicit `public` (Code-style: modules are default `private` with explicit public interfaces).**
   - `schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90` (no `private`, no `public`)
   - `schemes/pumas/micro_pumas_ccpp_dimensions_post.F90` (has `public` but no `private`)
   - `schemes/pumas/pumas_pre_main.F90` (no `private`, no `public`)

### Low priority / for human review

8. **Comment grammar/typo (Code-review rule 9) — `schemes/pumas/pumas_pre_main.F90` header.** "The hope that there will eventually be one set of portable interstitials..." is a sentence fragment (missing "is" — "The hope *is* that..."), and "intersititals" is misspelled (→ "interstitials").

9. **`schemes/sima_diagnostics/pumas_diagnostics.F90-keep_for_run` (2150 lines added).** This file has a non-`.F90` extension and appears to be a leftover working/backup copy committed by accident (GitHub can't even diff it). It falls under the diagnostics work you've flagged as out-of-review, but a `-keep_for_run` file almost certainly should not be committed to the repo — please confirm/remove.

10. *(Optional, Code-review rule 3 — `pure`.)* The pure data-reshaping converters do no I/O and have no side effects, so they are candidates for the `pure` keyword: `micro_pumas_ccpp_dimensions_pre_*`, `micro_pumas_ccpp_dimensions_post_run`, `micro_pumas_ccpp_optics_limiter_run`, and `clamp_number_concentrations_run`. (`pumas_post_main_run` calls external `size_dist_param_*` routines, so it likely can't be `pure`.) Flagging as optional since it isn't universal practice for CCPP entry points here.

11. *(Minor, Code-style modern syntax.)* `schemes/utilities/clamp_number_concentrations.F90:~112` uses the old array-constructor `(/ ... /)`; prefer `[ ... ]`.

### Notes (not blocking)

- The `effective_radius_of_stratiform_cloud_{liquid_water,ice}_particle` → `..._droplet`/`..._crystal` and `effective_diameter_..._particle_for_radiation` → `..._crystal_for_radiation` renames are applied consistently across `rk_stratiform.{F90,meta}`, `rrtmgp_{lw,sw}_cloud_optics.meta`, and `rk_stratiform_diagnostics.{F90,meta}`, with units unchanged (`um`). No rule-14 (same `standard_name`, differing `units`) conflicts were found in the modified `.meta` files.
- The new `mass_number_concentration_*_{droplets,drops,crystals,particles}_in_moist_air_and_condensed_water` names in `initialize_constituents.F90` are consistent with their use in `clamp_number_concentrations.F90`. As a reminder (`AGENTS.md`), any newly coined standard names should be explicitly called out in the PR description for SE review.
- For later: `pumas_diagnostics.F90` and `pumas_diagnostics_calling.F90` also carry module-level `save` statements (rule 1) — worth addressing when those files come up for review.

schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90
```diff
@@ -0,0 +1,244 @@
...
+                                                       !cloud physics is applied (1 to nlev)
+
+    !PUMAS dimensions/parameters:
+    integer,         intent(out) :: micro_ncol         !Number of horizontal microphysics columns (count)
+    integer,         intent(out) :: micro_nlev         !Number of microphysics vertical layers (count)
+    integer,         intent(out) :: micro_nlevp1       !Number of microphysics vertical interfaces (count)
+
+    !CCPP error handling:
+    character(len=*), intent(out) :: errmsg
+    integer,            intent(out) :: errcode
```
nusbaume: Fix indentation here to better align with the code above?


```suggestion
    integer,          intent(out) :: errcode
```

schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90
```diff
@@ -0,0 +1,244 @@
...
+    integer,         intent(in) :: nlev
+    integer,         intent(in) :: nlevp1
+    integer,         intent(in) :: trop_cloud_top_lev  !Index of the top model level for which
+                                                       !cloud physics is applied (1 to nlev)
+
+    !PUMAS dimensions/parameters:
+    integer,         intent(in) :: micro_ncol          !Number of horizontal microphysics columns (count)
+    integer,         intent(in) :: micro_nlev          !Number of microphysics vertical layers (count)
+    integer,         intent(in) :: micro_nlevp1        !Number of microphysics vertical interfaces (count)
+    integer,         intent(in) :: micro_dust_nbins    !Number of dust bins
```
nusbaume: It looks like most of these variables aren't used in this subroutine, so I _think_ this can be cleaned up to be the following:


```suggestion
    !Host model dimensions/parameters:
    integer,         intent(in) :: ncol
    integer,         intent(in) :: trop_cloud_top_lev  !Index of the top model level for which
                                                       !cloud physics is applied (1 to nlev)
```

schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90
```diff
@@ -0,0 +1,244 @@
+module micro_pumas_ccpp_dimensions_pre
+
+  use pumas_kinds,       only: pumas_r8=>kind_r8
+  use ccpp_kinds,        only: kind_phys
```
nusbaume: This should have been caught by the AI, but can we move these `use` statements down to the relevant subroutine level?

schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
```diff
@@ -0,0 +1,670 @@
...
+  long_name = cloud liquid wrt moist air and condensed water of new state
+  units = kg kg-1
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  type = real
+  kind = kind_phys
+  intent = in
+  advected = true
+[pumas_cldliq]
+  standard_name = pumas_cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water
+  long_name = microphysics cloud liquid wrt moist air and condensed water of new state
```
nusbaume: I might add "mixing ratio" to the long name here:


```suggestion
  long_name = microphysics cloud liquid mixing ratio wrt moist air and condensed water of new state
```

schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
```diff
@@ -0,0 +1,670 @@
...
+  long_name = tendency of cloud liquid droplet number concentration due to immersion freezing
+  units = cm-3
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  type = real
+  kind = kind_phys
+  intent = in
+[pumas_frzimm]
+  standard_name = tendency_of_pumas_cloud_liquid_droplet_number_concentration_due_to_immersion_freezing
+  long_name = microphysics tendency of cloud liquid droplet number concentration due to immersion freezing
+  units = cm-3
```
nusbaume, line 571: I am not sure why I didn't catch this earlier, but essentially by definition a variable that is a `tendency_of_X` must have a `s-1` term in the units, which it looks like the three `frz` variables listed here don't have.

I wouldn't worry about it for this PR, but it might be good to create an issue pointing out this discrepency, as it could potentially indicate either a science bug or an incorrect description of what the variable actually is.

schemes/pumas/pumas_pre_main.F90
```diff
@@ -0,0 +1,58 @@
+! CCPP interstitial schemes needed for the PUMAS cloud microphysics package.
+
+!The hope is that there will eventually be one set of portable interstitials and possibly another set
+!of host-specific interstitials (if needed) above the portable layer.
+
+module pumas_pre_main
+
+   use ccpp_kinds,        only: kind_phys
```
nusbaume: Move this `use` statement down to the relevant subroutine level?

schemes/pumas/pumas_pre_main.F90
```diff
@@ -0,0 +1,58 @@
...
+       remove_supersat = .false.
+     else
+       remove_supersat = .true.
+     end if
+
+   end subroutine pumas_pre_main_init
+
+  !> \section arg_table_pumas_pre_main_timestep_init Argument Table
+  !! \htmlinclude pumas_pre_main_timestep_init.html
+   subroutine pumas_pre_main_timestep_init( errmsg, errcode)
```
nusbaume: I might remove this subroutine, as it doesn't appear to be doing anything.

schemes/pumas/pumas_pre_main_namelist.xml
```diff
@@ -0,0 +1,602 @@
...
+
+<entry id="stochastic_emulated_filename_quantile">
+  <category>microphys</category>
+  <group>micro_mg_nl</group>
+  <standard_name>filename_of_emulated_stochastic_collection_quantiles</standard_name>
+  <units>none</units>
+  <type>char*256</type>
+  <desc>emulated stochastic collection filename for quantiles</desc>
+  <values>
+    <value>"${SRCROOT}2/src/atmos_phys/schemes/pumas/pumas/KBARF_tau_kernel.dat"</value>
```
nusbaume: I don't think the number 2 is supposed to be here:


```suggestion
    <value>"${SRCROOT}/src/atmos_phys/schemes/pumas/pumas/KBARF_tau_kernel.dat"</value>
```

schemes/pumas/pumas_pre_main_namelist.xml
```diff
@@ -0,0 +1,602 @@
...
+  <category>microphys</category>
+  <group>micro_mg_nl</group>
+  <standard_name>prescribed_cloud_droplet_number_concentration</standard_name>
+  <units>m-3</units>
+  <type>real</type>
+  <kind>kind_phys</kind>
+  <desc>droplet concentration constant for MG microphysics</desc>
+  <values>
+    <value>50000000.</value>
+    <value micro_mg_version="2">8</value>
```
nusbaume: It looks like this is a reference to the wrong "ncnst", so I would remove it (especially because a value of `8` would be wildly out-of-range compared to the real world).


```suggestion
```

schemes/microp_aero/ccpp/hetfrz_classnuc_stub.F90
```diff
@@ -0,0 +1,56 @@
+! Stub for heterogeneous freezing by classical nucleation theory (hetfrz_classnuc).
```
nusbaume, line 1: Just a reminder to add all of the `hetfrz_classnuc_stub.*` code to the PR description!

schemes/microp_aero/ccpp/hetfrz_classnuc_stub.F90
```diff
@@ -0,0 +1,56 @@
...
+!
+! This scheme exists to read the use_hetfrz_classnuc namelist flag and expose it
+! via its CCPP standard name so downstream schemes (e.g. nucleate_ice_ccpp,
+! eventually PUMAS) can see the flag. It performs no heterogeneous-freezing
```
nusbaume: I might replace `eventually PUMAS` with just `PUMAS`, as I believe it will be wired up correctly once this PR is merged?


```suggestion
! PUMAS) can see the flag. It performs no heterogeneous-freezing
```

schemes/microp_aero/ccpp/hetfrz_classnuc_stub_namelist.xml
```diff
@@ -0,0 +1,102 @@
...
+  <entry id="use_hetfrz_classnuc">
+    <type>logical</type>
+    <category>microphys</category>
+    <group>hetfrz_classnuc_nl</group>
+    <standard_name>do_heterogeneous_ice_nucleation</standard_name>
+    <units>flag</units>
+    <desc>
+      Switch to turn on heterogeneous freezing by classical nucleation theory.
+      Currently a stub in CAM-SIMA: the flag is propagated to downstream schemes
+      (nucleate_ice_ccpp, eventually PUMAS) but no hetfrz computation is performed.
```
nusbaume: Remove `eventually` here?


```suggestion
      (nucleate_ice_ccpp, PUMAS) but no hetfrz computation is performed.
```

test/test_schemes/clubb_stub.F90
```diff
@@ -0,0 +1,23 @@
...
+public :: clubb_stub_register
+
+contains
+
+!> \section arg_table_clubb_stub_register  Argument Table
+!! \htmlinclude clubb_stub_register.html
+subroutine clubb_stub_register(do_clubb_sgs, errmsg, errcode)
+    ! Dummy variables
+    logical, intent(out)            :: do_clubb_sgs
+    character(len=512), intent(out) :: errmsg
```
nusbaume: I would change this to `len=*`:


```suggestion
    character(len=*), intent(out) :: errmsg
```

test/test_schemes/initialize_constituents.F90
```diff
@@ -56,12 +56,12 @@ subroutine initialize_constituents_register(constituents, errmsg, errcode)
...
-    character(len=75), parameter :: water_species_number_std_names(5) = &
-       ['mass_number_concentration_of_cloud_liquid_wrt_moist_air_and_condensed_water', &
-        'mass_number_concentration_of_rain_wrt_moist_air_and_condensed_water        ', &
-        'mass_number_concentration_of_ice_wrt_moist_air_and_condensed_water         ', &
-        'mass_number_concentration_of_snow_wrt_moist_air_and_condensed_water        ', &
-        'mass_number_concentration_of_graupel_wrt_moist_air_and_condensed_water     ']
+    character(len=*), parameter :: water_species_number_std_names(*) = &
+       [character(len=128) :: 'mass_number_concentration_of_cloud_liquid_water_droplets_in_moist_air_and_condensed_water', &
+                              'mass_number_concentration_of_rain_drops_in_moist_air_and_condensed_water', &
+                              'mass_number_concentration_of_ice_wrt_moist_air_and_condensed_water', &
```
nusbaume: Should this be `cloud_ice_water_crystals` here?


```suggestion
                              'mass_number_concentration_of_cloud_ice_water_crystals_wrt_moist_air_and_condensed_water', &
```

test/test_suites/suite_pumas.xml
```diff
@@ -0,0 +1,33 @@
...
+  <group name="physics_before_coupler">
+
+    <scheme>initialize_constituents</scheme>
+    <scheme>cloud_fraction_options</scheme>
+    <scheme>compute_cloud_fraction_two_moment</scheme>
+    <scheme>to_be_ccppized_temporary</scheme>
+    <scheme>clubb_stub</scheme>
+    <scheme>hetfrz_classnuc_stub</scheme>
+
+    <!-- Read in PUMAS namelist variables (init only) -->
```
nusbaume, line 13: Just double-checking if the plan is to add PUMAS to the `suite_cam5.xml` and `suite_cam7.xml` SDFs once diagnostics are complete?

test/test_suites/suite_pumas.xml
```diff
@@ -0,0 +1,33 @@
...
+    <!-- Apply tendencies -->
+    <scheme>apply_constituent_tendencies</scheme>
+    <scheme>clamp_number_concentrations</scheme>
+    <!-- Recompute size distribution quantities for radiation from the updated
+         constituents; needs the pre-heating air temperature, so it must run
+         before apply_heating_rate -->
+    <scheme>pumas_post_main</scheme>
+    <scheme>micro_pumas_ccpp_optics_limiter</scheme>
+    <scheme>apply_heating_rate</scheme>
+    <scheme>qneg</scheme>
```
nusbaume: Can we move `qneg` to right after `clamp_number_concentrations`?  This would be better algorithmically (as we don't want any scheme to see negative constituent values), but if that causes answer changes then we can just make an issue to fix it later.

schemes/utilities/clamp_number_concentrations.F90
```diff
@@ -0,0 +1,126 @@
...
+
+  ! Clamp bounds
+  real(kind_phys), parameter :: qmin = 1.0e-12_kind_phys ! minimum number concentration [kg-1]
+  real(kind_phys), parameter :: qmax = 1.0e10_kind_phys  ! maximum number concentration [kg-1]
+
+  ! Standard names for each species (order matches ix_ variables above)
+  character(len=*), parameter :: std_names(num_species) = &
+    [character(len=128) :: 'mass_number_concentration_of_cloud_liquid_water_droplets_in_moist_air_and_condensed_water', &
+                           'mass_number_concentration_of_rain_drops_in_moist_air_and_condensed_water', &
+                           'mass_number_concentration_of_ice_wrt_moist_air_and_condensed_water', &
```
nusbaume: I believe this should say `cloud_ice_water_crystals`:


```suggestion
                           'mass_number_concentration_of_cloud_ice_water_crystals_wrt_moist_air_and_condensed_water', &
```

schemes/utilities/clamp_number_concentrations.F90
```diff
@@ -0,0 +1,126 @@
...
+                           'mass_number_concentration_of_snow_crystals_in_moist_air_and_condensed_water']
+
+contains
+
+!> \section arg_table_clamp_number_concentrations_init Argument Table
+!! \htmlinclude clamp_number_concentrations_init.html
+  subroutine clamp_number_concentrations_init(const_props, errmsg, errflg)
+
+    use ccpp_constituent_prop_mod, only: ccpp_constituent_prop_ptr_t
+    use ccpp_const_utils,          only: ccpp_const_get_idx
```
nusbaume: I would use the CCPP-framework provided utility here:


```suggestion
    use ccpp_scheme_utils,          only: ccpp_constituent_index
```

This would also mean that you no longer need to bring in the `const_props` object as well.

schemes/utilities/clamp_number_concentrations.F90
```diff
@@ -0,0 +1,126 @@
...
+
+    type(ccpp_constituent_prop_ptr_t), &
+                        intent(in)  :: const_props(:)
+    character(len=512), intent(out) :: errmsg
```
nusbaume: This should be `len=*`:


```suggestion
    character(len=*), intent(out) :: errmsg
```

schemes/utilities/clamp_number_concentrations.F90
```diff
@@ -0,0 +1,126 @@
...
+    integer,            intent(out) :: errflg
+
+    ! Local variables
+    integer :: ix_species(num_species)
+    integer :: n
+    character(len=512) :: local_errmsg
+    integer            :: local_errflg
```
nusbaume: I don't believe these variables are needed if you use the framework-provided `ccpp_constituent_index` routine.

schemes/utilities/clamp_number_concentrations.F90
```diff
@@ -0,0 +1,126 @@
...
+    ! Look up constituent indices. A missing species is skipped
+    do n = 1, num_species
+      call ccpp_const_get_idx(const_props, trim(std_names(n)), &
+                              ix_species(n), local_errmsg, local_errflg)
+      if (local_errflg /= 0) then
+        ! Constituent not found — mark as inactive, reset error
+        ix_species(n) = -1
+        local_errflg  = 0
+        local_errmsg  = ''
+      end if
```
nusbaume: If you use `ccpp_constituent_index` here then I believe this can be simplified to the following:


```suggestion
      call ccpp_constituent_index(trim(std_names(n)), ix_species(n))
      if (ix_species(n) < 0) then
        ! Constituent not found — mark as inactive
        ix_species(n) = -1
      end if
```

schemes/utilities/clamp_number_concentrations.F90
```diff
@@ -0,0 +1,126 @@
...
+    ncol, pver, dt, &
+    const_q, const_tend, &
+    errmsg, errflg)
+
+    integer,            intent(in)    :: ncol
+    integer,            intent(in)    :: pver
+    real(kind_phys),    intent(in)    :: dt                  ! physics timestep [s]
+    real(kind_phys),    intent(inout) :: const_q(:, :, :)    ! constituent mixing ratios
+    real(kind_phys),    intent(inout) :: const_tend(:, :, :) ! constituent tendencies
+    character(len=512), intent(out)   :: errmsg
```
nusbaume: This should be `len=*`:


```suggestion
    character(len=*), intent(out)   :: errmsg
```

schemes/utilities/clamp_number_concentrations.meta
```diff
@@ -0,0 +1,71 @@
...
+
+[ccpp-arg-table]
+  name  = clamp_number_concentrations_init
+  type  = scheme
+[ const_props ]
+  standard_name = ccpp_constituent_properties
+  units = none
+  type = ccpp_constituent_prop_ptr_t
+  dimensions = (number_of_ccpp_constituents)
+  intent = in
```
nusbaume: Just a reminder that if you use `ccpp_constituent_index` then this metadata entry can be removed (as it is no longer needed as a subroutine argument).

schemes/pumas/micro_pumas_ccpp_dimensions_post.F90
```diff
@@ -0,0 +1,493 @@
...
+    ! vertical layer dimension
+    integer, intent(in) :: nlev
+    ! microphysics vertical layer dimension
+    integer, intent(in) :: micro_nlev
+    ! vertical interface dimension
+    integer, intent(in) :: nlevp1
+    ! microphysics vertical interface dimension
+    integer, intent(in) :: micro_nlevp1
+    ! index of the top model level for which cloud physics is applied (1 to nlev)
+    integer, intent(in) :: trop_cloud_top_lev
```
nusbaume, line 49: I think most of these subroutine arguments aren't needed:


```suggestion
    integer, intent(in) :: ncol

    ! index of the top model level for which cloud physics is applied (1 to nlev)
    integer, intent(in) :: trop_cloud_top_lev
```

schemes/pumas/micro_pumas_ccpp_dimensions_post.F90
```diff
@@ -0,0 +1,493 @@
...
+
+    character(len=*), intent(out) :: errmsg  !PUMAS/CCPP error message (none)
+    integer,            intent(out) :: errcode !CCPP error code (1)
+
+    ! Set error variables
+    errcode = 0
+    errmsg = ''
+
+    ! Copy PUMAS outputs (levels 1:micro_nlev) back onto host levels trop_cloud_top_lev:nlev.
+    qcsinksum_rate1ord(:,trop_cloud_top_lev:)   = pumas_qcsinksum_rate1ord(:ncol,:)
```
nusbaume, line 341: I am not 100% sure it's required, but I believe we can avoid compiler warnings (or at least comments) if we make the kind conversion explicit here and below:


```suggestion
    qcsinksum_rate1ord(:,trop_cloud_top_lev:)   = real(pumas_qcsinksum_rate1ord(:ncol,:), kind_phys)
```

schemes/pumas/micro_pumas_ccpp_dimensions_post.F90
```diff
@@ -0,0 +1,493 @@
+module micro_pumas_ccpp_dimensions_post
+
+  use pumas_kinds,       only: pumas_r8=>kind_r8
+  use ccpp_kinds,        only: kind_phys
```
nusbaume: Can we move these `use` statements down to the subroutine level?

schemes/pumas/micro_pumas_ccpp_dimensions_post.meta
```diff
@@ -0,0 +1,1206 @@
+[ccpp-table-properties]
+  name = micro_pumas_ccpp_dimensions_post
+  type = scheme
```
nusbaume, line 3: Don't we need a `dependencies` and `kind_spec` attribute here?


```suggestion
  type = scheme
  dependencies = pumas/pumas_kinds.F90
  kind_spec = pumas_kinds:pumas_r8=>kind_r8
```

schemes/pumas/micro_pumas_ccpp_dimensions_post.meta
```diff
@@ -0,0 +1,1206 @@
...
+[pumas_numliq_tend]
+  standard_name = tendency_of_pumas_mass_number_concentration_of_cloud_liquid_water_droplets_in_moist_air_and_condensed_water
+  long_name = microphysics mass number concentration of cloud liquid water wrt moist air and condensed water
+  units = kg-1 s-1
+  dimensions = (microphysics_horizontal_loop_extent, microphysics_vertical_layer_dimension)
+  type = real
+  kind = pumas_r8
+  intent = in
+[numice_tend]
+  standard_name = tendency_of_mass_number_concentration_of_ice_wrt_moist_air_and_condensed_water
```
nusbaume: I believe this should say `cloud_ice_water_crystals`:


```suggestion
  standard_name = tendency_of_mass_number_concentration_of_cloud_ice_water_crystals_wrt_moist_air_and_condensed_water
```

schemes/pumas/pumas_post_main.F90
```diff
@@ -0,0 +1,157 @@
...
+                max(mincld, real(strat_cldfrc(:ncol,:), pumas_r8)), 0.005_pumas_r8)
+    ncic(:,:)   =    real(numliq(:ncol,:), pumas_r8) / &
+                max(mincld, real(strat_cldfrc(:ncol,:), pumas_r8))
+
+    mu(:,:)      = 0._pumas_r8
+    lambdac(:,:) = 0._pumas_r8
+    do k = trop_cloud_top_lev, nlev
+      call size_dist_param_liq(mg_liq_props, icwmrst(:,k), ncic(:,k), rho(:,k), &
+                               mu(:,k), lambdac(:,k), ncol)
+    end do
```
nusbaume, line 110: There _might_ be a science bug here, in that the `rho` is calculated using the dry air gas constant and regular temperature (but the full pressure), and is then being used to convert the value of various mass number concentration variables (like `ncic`), which is with respect to the mass of moist air and condensed water.  I obviously wouldn't worry about fixing it here, but might be worth opening an issue so that we can double-check it in the future?

Also pinging @PeterHjortLauritzen in case I am totally wrong here.

schemes/pumas/pumas_post_main.meta
```diff
@@ -0,0 +1,196 @@
+[ccpp-table-properties]
+  name = pumas_post_main
+  type = scheme
+  dependencies = pumas/micro_pumas_utils.F90
```
nusbaume, line 4: Do we need the kind-related dependencies here as well?

```suggestion
  type = scheme
  dependencies = pumas/micro_pumas_utils.F90
  dependencies = pumas/pumas_kinds.F90
  kind_spec = pumas_kinds:pumas_r8=>kind_r8
```

schemes/pumas/pumas_post_main.meta
```diff
@@ -0,0 +1,196 @@
...
+  standard_name = cloud_ice_mixing_ratio_wrt_moist_air_and_condensed_water
+  long_name = cloud ice mixing ratio wrt moist air and condensed water
+  units = kg kg-1
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  type = real
+  kind = kind_phys
+  intent = in
+  advected = true
+[numice]
+  standard_name = mass_number_concentration_of_ice_wrt_moist_air_and_condensed_water
```
nusbaume: I believe this should be `cloud_ice_water_crystals` here:


```suggestion
  standard_name = mass_number_concentration_of_cloud_ice_water_crystals_wrt_moist_air_and_condensed_water
```

nusbaume (changes requested):
Thanks for getting all of the required interstitials set up @cacraigucar and @jimmielin!  I have some comments and change requests, but I believe many should either be easy to fix, or can just be pushed off onto future work issues (including the standard name requests).  Of course if you have any questions or concerns just let me know!

schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
jimmielin: Just a note to feel free to assign this issue to me as I will deal with the units when CCPPizing hetfrz_classnuc eventually, beyond the MAM work.

schemes/pumas/pumas_post_main.F90
jimmielin: Just a note that I think this bug belongs to ESCOMP/PUMAS.

The post-interstitial rho is consistent with the one inside `micro_pumas_v1.F90`
```fortran
        rho(i,k) = p(i,k)/(r*t(i,k))
```
`p` here is also `pmid`, `r` is `rair`

The `rho` cancels out in `des` and `degrau`, I think, so it's just `pgam` and `lamc`.

If a fix is made it should be in PUMAS then the downstream interstitials are adjusted to remain consistent. This would introduce answer changes everywhere, though.

jimmielin: After discussing with @cacraigucar, I added https://github.com/ESCOMP/PUMAS/issues/77 since a fix would have to be at least coordinated between CAM and SIMA

schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90
cacraigucar: done

schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90
cacraigucar: done

schemes/pumas/micro_pumas_ccpp_dimensions_pre.F90
cacraigucar: removed

schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
```diff
@@ -0,0 +1,670 @@
...
+  standard_name = cloud_ice_mixing_ratio_wrt_moist_air_and_condensed_water
+  long_name = cloud ice mixing ratio wrt moist air and condensed water of new state
+  units = kg kg-1
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  type = real
+  kind = kind_phys
+  intent = in
+  advected = true
+[pumas_cldice]
+  standard_name = pumas_cloud_ice_mixing_ratio_wrt_moist_air_and_condensed_water
```
cacraigucar, line 157: changed to mass_number_concentration_of_cloud_ice_water_crystals_in_moist_air_and_condensed_water

schemes/pumas/pumas_pre_main.F90
cacraigucar: done

schemes/pumas/pumas_pre_main.F90
cacraigucar: removed

schemes/pumas/pumas_pre_main_namelist.xml
cacraigucar: removed

schemes/pumas/pumas_pre_main_namelist.xml
cacraigucar: Not sure where that came from!

schemes/microp_aero/ccpp/hetfrz_classnuc_stub.F90
cacraigucar: Done

schemes/microp_aero/ccpp/hetfrz_classnuc_stub.F90
cacraigucar: removed

schemes/microp_aero/ccpp/hetfrz_classnuc_stub_namelist.xml
cacraigucar: removed

test/test_schemes/clubb_stub.F90
cacraigucar: changed

test/test_schemes/initialize_constituents.F90
cacraigucar: Actually changed to: mass_number_concentration_of_cloud_ice_water_crystals_in_moist_and_condensed_water

test/test_suites/suite_pumas.xml
cacraigucar: moved without changing answers

schemes/utilities/clamp_number_concentrations.F90
cacraigucar: Done with the correct name

schemes/pumas/pumas_post_main.F90
PeterHjortLauritzen: > There _might_ be a science bug here, in that the `rho` is calculated using the dry air gas constant and regular temperature (but the full pressure), and is then being used to convert the value of various mass number concentration variables (like `ncic`), which is with respect to the mass of moist air and condensed water. I obviously wouldn't worry about fixing it here, but might be worth opening an issue so that we can double-check it in the future?
> 
> Also pinging @PeterHjortLauritzen in case I am totally wrong here.

Yes you are correct: ```rho``` in the code above is not moist density since the temperature is not virtual:

The thermodynamically consistent moist density is ```pmid/(rair*Tv)``` so ```rho``` is biased high by roughly 0.61*q — order 1% in a warm boundary layer, well under 0.1% in the upper troposphere where most of the ice/liquid size-distribution work happens.

schemes/utilities/clamp_number_concentrations.F90
jimmielin: Fixed in 8e28317d

schemes/utilities/clamp_number_concentrations.F90
jimmielin: Fixed in 8e28317d

schemes/utilities/clamp_number_concentrations.F90
jimmielin: Fixed in 8e28317d

schemes/utilities/clamp_number_concentrations.F90
jimmielin: Fixed in 8e28317d

schemes/utilities/clamp_number_concentrations.F90
jimmielin: Fixed in 8e28317d

schemes/utilities/clamp_number_concentrations.meta
jimmielin: Fixed in 8e28317d

schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
```diff
@@ -183,7 +138,7 @@
   advected = true
 [pumas_cldliq]
   standard_name = pumas_cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water
-  long_name = microphysics cloud liquid wrt moist air and condensed water of new state
+  long_name = microphysics cloud mixing ratio liquid wrt moist air and condensed water of new state
```
nusbaume: I would put `mixing ratio` after `liquid` here.

test/test_suites/suite_pumas.xml
nusbaume: For `suite_cam5.xml`, I believe you can put the PUMAS code-block right after the `compute_cloud_fraction` scheme on line 98 (or basically right before the `sima_state_diagnostics`.

For `suite_cam7.xml`, I believe you can put the PUMAS code-block right after the `Vertical diffusion (CLUBB)` comment on line 117, or right before the `Gravity wave drag` comment.

schemes/pumas/micro_pumas_ccpp_dimensions_post.F90
nusbaume: Looks like this request was missed?

schemes/pumas/pumas_post_main.meta
nusbaume: Looks like this request was missed as well?

nusbaume (changes requested):
Thanks @cacraigucar for implementing my requests!  It looks like there were a few missed requests, but otherwise we are getting super-close!

schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
nusbaume: Let's make an issue for this @cacraigucar

schemes/pumas/pumas_post_main.F90
nusbaume: Let's make an issue for this @cacraigucar.

schemes/pumas/pumas_post_main.meta
cacraigucar: done

schemes/pumas/pumas_post_main.F90
cacraigucar: @jimmielin has already opened ESCOMP/PUMAS#77 (see above)

schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
cacraigucar: done

schemes/pumas/micro_pumas_ccpp_dimensions_post.F90
cacraigucar: done

cacraigucar (commented):
I thought you may have mentioned two places that needed issues opened, but I only saw one.  I believe I've addressed everything

schemes/pumas/micro_pumas_ccpp_dimensions_pre.meta
nusbaume: I went ahead and created the new issue, which can be found here:  https://github.com/ESCOMP/atmospheric_physics/issues/434

nusbaume (approved):
Thanks @cacraigucar!  I went ahead and manually made a few last code edits (I hope that was ok!), but everything looks good to me now.  Thanks again for helping to bring in this major physics scheme!

schemes/pumas/pumas_pre_main.meta
```diff
@@ -0,0 +1,47 @@
...
+  type = scheme
+[ do_clubb_sgs ]
+  standard_name = do_clubb
+  long_name = Flag indicating whether CLUBB is being used in the suite or not
+  units = flag
+  type = logical
+  dimensions = ()
+  intent = in
+[spat_vary_accre_enhan_in]
+  standard_name = spatial_varying_accretion_enhancement_factor
```
jimmielin: I believe this should be `accretion_enhancement_factor` to match up with the registry standard name. Otherwise this standard name is not found anywhere else except in this file.

```suggestion
  standard_name = accretion_enhancement_factor
```

This shouldn't be answer changing as it is hardcoded in 1 in CAM anyway as the comments say.

schemes/pumas/pumas_post_main.F90
```diff
@@ -0,0 +1,157 @@
...
+    ! Ice effective radius is recomputed here from the post-microphysics
+    ! grid-mean in-cloud ice, NOT from the raw per-substep value PUMAS returns.
+    ! PUMAS uses the total stratiform fraction for ice cloud (icecldf = ast).
+    icimrst(:,:) = min(real(cldice(:ncol,:), pumas_r8) / &
+                max(mincld, real(strat_cldfrc(:ncol,:), pumas_r8)), 0.005_pumas_r8)
+    niic(:,:)   =    real(numice(:ncol,:), pumas_r8) / &
+                max(mincld, real(strat_cldfrc(:ncol,:), pumas_r8))
+
+    rei(:,:) = 25._pumas_r8
+    do k = 1, nlev
```
jimmielin: Looking at the CAM equivalent in `micro_pumas_cam.F90:3181` this should be

```suggestion
    do k = trop_cloud_top_lev, nlev
```

There should be no answer impact to CAM-SIMA since right now the model top is too low for it to make a difference but this would be buggy later when we introduce waccm

schemes/pumas/pumas_pre_main_namelist.xml
```diff
@@ -0,0 +1,601 @@
...
+
+<entry id="precip_fall_corr">
+  <category>microphys</category>
+  <group>micro_mg_nl</group>
+  <standard_name>do_precipitation_fall_speed_correction</standard_name>
+  <units>flag</units>
+  <type>logical</type>
+  <desc>ensure non-zero precipitation fallspeed for PUMAS microphysics</desc>
+  <values>
+    <value>.false.</value>
```
jimmielin:
```suggestion
    <value>.false.</value>
    <value phys_suite="cam7">.true.</value>
```

suites/suite_cam5.xml
```diff
@@ -6,12 +6,12 @@
...
       Shallow convection      UW         (not implemented)
       Macrophysics            Park       (not implemented)
-      Microphysics            PUMAS      (not implemented)
+      Microphysics            PUMAS
           * note - original was MG1.
       Radiation               RRTMGP
       Chemistry               None       (not implemented)
       Vertical Diffusion      UW (Bretherton-Park)
-        Orographic Stress     Beljaars
+      Orographic Stress       Beljaars
```
jimmielin: Orographic stress is supposed to be indented to below vertical diffusion as it provides the stresses to it; please revert the change:

```suggestion
        Orographic Stress     Beljaars
```

suites/suite_cam5.xml
```diff
@@ -97,6 +97,34 @@
     -->
     <scheme>compute_cloud_fraction</scheme>
 
+    <!-- cloud microphysics (PUMAS) -->
+
+    <scheme>initialize_constituents</scheme>
+    <scheme>cloud_fraction_options</scheme>
+    <scheme>compute_cloud_fraction_two_moment</scheme>
+    <scheme>to_be_ccppized_temporary</scheme>
```
jimmielin: These lines are duplicates of the initialization above and `initialize_constituents` is not needed, so they should not be added here

```suggestion
```

suites/suite_cam5.xml
```diff
@@ -97,6 +97,34 @@
...
+    <scheme>clubb_stub</scheme>
```
jimmielin: This shouldn't be here as CAM5 does not use CLUBB and `clubb_stub` hardcodes `do_clubb_sgs = .true.`

Let's just remove this for now and it can be replaced with a registry flag in the future.

```suggestion
```

jimmielin (commented):
Thanks @cacraigucar. I have a few final change requests, I tried to keep to the minimal changes since we want to get this to the finish line, so I kept only the priority ones. These should not have any answer impact on CAM (since it's only interstitials and SDFs) and should not change results on CAM-SIMA.

After PUMAS round 3 is merged, I'll open a follow-up PR with other lower priority fixes.

jimmielin (approved):

Issue #339, opened 2025-12-09
Unify constituent minimum values across schemes


peverwhee:
When run-time constituents are registered in a scheme with an `instantiate` call, a minimum value is supplied. If the same constituent is registered by multiple schemes and that minimum value differs, the model will error.

Among possibly other options, we could:
1. Create a lookup method that is accessible by any scheme that would enable you to get the "proper" minimum value and set it.
2. Initialize every constituent with min value = 0 and then have a scheme that calls `set_minimum` for each registered constituent.

related to #332

issue noted by @jimmielin in #327

PR #367, opened 2026-03-05
qmin changes: minimum constituent values for chemistry/aerosol consistent with CAM

jimmielin:
Originator(s): @jimmielin

- add `ccpp_chem_utils::chem_constituent_qmin` to set `qmin` value for chemistry species consistent with CAM (values are from `chem_register`).
- update `initialize_constituents` scheme to use these "correct" `qmin` values.
- partial work on https://github.com/ESCOMP/atmospheric_physics/issues/339 (need to update other register calls in `atmos_phys`, if any, with chemistry constituents)

```
A       to_be_ccppized/ccpp_chem_utils.F90
  - add chem_constituent_qmin

M       test/test_schemes/initialize_constituents.F90
  - use correct qmin values consistent with CAM
```

Not for current schemes, but will be necessary for future (CAM5+ schemes) that are sensitive to qmin

### DIFF
--- a/test/test_schemes/initialize_constituents.F90
+++ b/test/test_schemes/initialize_constituents.F90
@@ -24,15 +24,15 @@ subroutine initialize_constituents_register(constituents, errmsg, errcode)
     integer :: num_variables
     integer :: ierr
     integer :: var_index
-    integer :: split_index
     integer :: constituent_index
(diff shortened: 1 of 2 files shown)


jimmielin:
will be part of CAM5 PBL scheme

PR #372, opened 2026-03-12
CCPPize prescribed volcanic aerosol (prescribe_volcaero)

### LINKED ISSUE #363: CCPPize prescribed_volcaero (prescribed volcanic aerosol)
`VOLC_MMR` for all aerosol models (for now, BAM) needs to be provided

Convert `prescribed_volcaero` ("manages reading and interpolation of prescribed volcanic aerosol") using the new tracer_data module in CAM-SIMA.

jimmielin:
Originator(s): @jimmielin 

- Closes #363 - this is the CAM-SIMA version of `prescribed_volcaero`; will NOT go back to CAM since the `tracer_data` interface is different
- Update `filename_for` standard names to `filename_of` for rule compliance

A       schemes/chemistry/prescribed_volcanic_aerosol_namelist.xml

```
A       schemes/chemistry/prescribed_volcanic_aerosol.F90
A       schemes/chemistry/prescribed_volcanic_aerosol.meta
  - new scheme
```

```
M       test/test_suites/suite_tracer_data_test.xml
  - add to existing tracer_data BAM test

M       schemes/chemistry/prescribed_aerosol_deposition_flux.meta
M       schemes/chemistry/prescribed_aerosol_deposition_flux_namelist.xml
M       schemes/chemistry/prescribed_aerosols.meta
M       schemes/chemistry/prescribed_aerosols_namelist.xml
  - fix filename_for -> filename_of to match standard names rules
```

New physics package for the tracer data test.

Updated snapshot; tested with CAM snapshot for correctness

### DIFF
--- a/schemes/chemistry/prescribed_ozone.meta
+++ b/schemes/chemistry/prescribed_ozone.meta
@@ -150,7 +150,7 @@
   intent = in
 [ boltz ]
   standard_name = boltzmann_constant
-  units = J K-1
+  units = J K-1 molecule-1
   type = real | kind = kind_phys
   dimensions = ()
   intent = in
--- a/schemes/chemistry/prescribed_ozone.meta
+++ b/schemes/chemistry/prescribed_ozone.meta
@@ -150,7 +150,7 @@
   intent = in
 [ boltz ]
   standard_name = boltzmann_constant
-  units = J K-1 molecule-1
+  units = J K-1
   type = real | kind = kind_phys
   dimensions = ()
   intent = in
--- a/schemes/chemistry/prescribed_volcanic_aerosol.F90
+++ b/schemes/chemistry/prescribed_volcanic_aerosol.F90
@@ -203,13 +203,9 @@ contains
       return
     end if
 
-    ! Register history fields
-    call history_add_field(volcaero_const_name, &
-         'prescribed volcanic aerosol dry mass mixing ratio', &
-         'lev', 'inst', 'kg kg-1')
-    call history_add_field(volcrad_const_name, &
-         'volcanic aerosol geometric-mean radius', &
-         'lev', 'inst', 'm')
+    ! Register history fields.
+    ! No longer need history output for the constituents because, well,
+    ! they are constituents.
     call history_add_field('VOLC_MASS', &
          'volcanic aerosol vertical mass path in layer', &
          'lev', 'inst', 'kg m-2')
@@ -352,8 +348,6 @@ contains
     columnmass(:ncol) = sum(volcmass(:ncol, :pver), 2)
 
     ! History output
-    call history_out_field(volcaero_const_name, constituents(:, :, mmr_idx))
-    call history_out_field(volcrad_const_name,  constituents(:, :, rad_idx))
     call history_out_field('VOLC_MASS',   volcmass(:, :))
     call history_out_field('VOLC_MASS_C', columnmass(:))
 
--- a/schemes/chemistry/prescribed_volcanic_aerosol.F90
+++ b/schemes/chemistry/prescribed_volcanic_aerosol.F90
@@ -73,7 +73,7 @@ contains
     errflg = 0
 
     ! Check if prescribed volcanic aerosols are enabled
-    if (prescribed_volcaero_file == 'NONE' .or. &
+    if (prescribed_volcaero_file == 'UNSET' .or. &
         len_trim(prescribed_volcaero_file) == 0) then
       has_prescribed_volcaero = .false.
       if (amIRoot) then
@@ -250,7 +250,7 @@ contains
     use ccpp_const_utils,          only: ccpp_const_get_idx
 
     ! dependency for unit string handling
-    use string_utils,              only: to_lower, GLC
+    use string_utils,              only: to_lower, get_last_significant_char
 
     integer,            intent(in)    :: ncol
     integer,            intent(in)    :: pver
@@ -305,7 +305,7 @@ contains
     if (errflg /= 0) return
 
     ! Determine unit conversion factor based on units in the input file
-    select case ( to_lower(trim(tracer_data_fields(1)%units(:GLC(tracer_data_fields(1)%units)))) )
+    select case ( to_lower(trim(tracer_data_fields(1)%units(:get_last_significant_char(tracer_data_fields(1)%units)))) )
     case ("molec/cm3", "/cm3", "molecules/cm3", "cm^-3", "cm**-3")
       ! Number density [molecules cm-3] -> MMR [kg kg-1]
       ! mmr = (M * 1e6 * k_B * T) / (M_air * p_dry)
--- a/schemes/chemistry/prescribed_volcanic_aerosol.F90
+++ b/schemes/chemistry/prescribed_volcanic_aerosol.F90
@@ -1,0 +1,362 @@
+! Manages reading and interpolation of prescribed volcanic aerosol concentrations.
+!
+! This module uses CCPP constituents (non-advected) to store prescribed volcanic aero
+! fields:
+! 1) volcanic aerosol mass mixing ratio (from prescribed dataset)
+! 2) geometric-mean wet aerosol radius (derived from mass)
+!
+! Based on original CAM version from: Francis Vitt
+module prescribed_volcanic_aerosol
+  use ccpp_kinds, only: kind_phys
+
+  ! CAM-SIMA host model dependency to read aerosol data
+  use tracer_data, only: trfile     ! data information and file read state
+  use tracer_data, only: trfld      ! tracer data container
+
+  implicit none
+  private
+
+  ! public CCPP-compliant subroutines
+  public :: prescribed_volcanic_aerosol_register
+  public :: prescribed_volcanic_aerosol_init
+  public :: prescribed_volcanic_aerosol_run
+
+  ! fields to store tracer_data state and information
+  type(trfld), pointer :: tracer_data_fields(:)
+  type(trfile)         :: tracer_data_file
+
+  ! module state variables
+  logical :: has_prescribed_volcaero = .false.
+
+  ! Constituent names
+  character(len=*), parameter :: volcaero_const_name = 'VOLC_MMR'
+  character(len=*), parameter :: volcrad_const_name  = 'VOLC_RAD_GEOM'
+
+  ! Molecular weight of volcanic aerosol species (sulfate) [g mol-1]
+  real(kind_phys), parameter :: molmass_volcaero = 47.9981995_kind_phys
+
+  ! WACCM-derived empirical coefficient relating mass concentration
+  ! to wet aerosol geometric-mean radius [m (kg m-3)^(-1/3)]
+  real(kind_phys), parameter :: radius_conversion = 1.9e-4_kind_phys
+
+  ! TODO: infrastructure for writing (and reading) tracer_data restart information.
+  ! see CAM/prescribed_volcrad_aero::{init,read,write}_prescribed_volcaero_restart
+  ! !!! Restarts will not be bit-for-bit without this !!!
+  ! TODO when SIMA implements restarts.
+
+contains
+
+  ! Register prescribed volcanic aerosol constituents.
+!> \section arg_table_prescribed_volcanic_aerosol_register Argument Table
+!! \htmlinclude prescribed_volcanic_aerosol_register.html
+  subroutine prescribed_volcanic_aerosol_register( &
+    amIRoot, iulog, &
+    prescribed_volcaero_file, &
+    volcaero_constituents, &
+    errmsg, errflg)
+
+    use ccpp_constituent_prop_mod, only: ccpp_constituent_properties_t
+
+    ! Input arguments:
+    logical,            intent(in)  :: amIRoot
+    integer,            intent(in)  :: iulog
+    character(len=*),   intent(in)  :: prescribed_volcaero_file          ! input filename from namelist
+
+    ! Output arguments:
+    type(ccpp_constituent_properties_t), allocatable, intent(out) :: volcaero_constituents(:)
+    character(len=*),   intent(out) :: errmsg
+    integer,            intent(out) :: errflg
+
+    character(len=*), parameter :: subname = 'prescribed_volcanic_aerosol_register'
+
+    errmsg = ''
+    errflg = 0
+
+    ! Check if prescribed volcanic aerosols are enabled
+    if (prescribed_volcaero_file == 'NONE' .or. &
+        len_trim(prescribed_volcaero_file) == 0) then
+      has_prescribed_volcaero = .false.
+      if (amIRoot) then
+        write(iulog,*) subname//': No prescribed volcanic aerosols specified'
+      end if
+      return
+    end if
+
+    has_prescribed_volcaero = .true.
+
+    ! Register two constituents: aerosol MMR and geometric-mean radius
+    allocate(volcaero_constituents(2), stat=errflg, errmsg=errmsg)
+    if (errflg /= 0) then
+      errmsg = subname // ": " // trim(errmsg)
+      return
+    end if
+
+    ! (1) Volcanic aerosol dry mass mixing ratio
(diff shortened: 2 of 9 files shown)


schemes/chemistry/prescribed_ozone.meta
```diff
@@ -150,7 +150,7 @@
   intent = in
 [ boltz ]
   standard_name = boltzmann_constant
-  units = J K-1
+  units = J K-1 molecule-1
```
nusbaume: I double-checked and the Boltzmann Constant is in fact just `J K-1`.  The code comments stating otherwise in CAM-SIMA and CESM_share are simply wrong.  So I would revert this change and update the relevant `physconst` comment in your CAM-SIMA PR (I can update the share code comment in my eventual water isotope PR).  Thanks!

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,356 @@
...
+
+    character(len=*), parameter :: subname = 'prescribed_volcanic_aerosol_register'
+
+    errmsg = ''
+    errflg = 0
+
+    ! Check if prescribed volcanic aerosols are enabled
+    if (prescribed_volcaero_file == 'UNSET' .or. &
+        len_trim(prescribed_volcaero_file) == 0) then
+      has_prescribed_volcaero = .false.
```
nusbaume: I don't think you need to set `has_prescribed_volcaero` here as it is already initialized to `.false.` at the module level.

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,356 @@
...
+         min_value         = 0.0_kind_phys, &
+         advected          = .false., &
+         water_species     = .false., &
+         mixing_ratio_type = 'dry', &
+         errcode           = errflg, &
+         errmsg            = errmsg)
+    if (errflg /= 0) return
+
+    ! (2) Volcanic aerosol geometric-mean radius (derived quantity)
+    call volcaero_constituents(2)%instantiate( &
```
nusbaume, line 109: I don't think it matters for this PR, but long-term I am not sure it is ideal to have the radius of the volcanic aerosols be its own constituent entry, because it is really not a constituent _per se_ (i.e. something that has "substance" or mass), but a property of the volcanic aerosols themselves.

If this is the only chemistry/aerosol scheme you know of in CAM that does this then I think we can ignore this concern, but if multiple aerosol configurations do it then it might be worth opening an issue so we have a place to start thinking about and discussing possible alternatives.

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,356 @@
...
+    ! Verify tracer_data is correctly initialized
+    if (.not. associated(tracer_data_fields)) then
+      errflg = 1
+      errmsg = subname//': tracer_data_fields not associated after trcdata_init'
+      return
+    end if
+
+    ! Register history fields.
+    ! No longer need history output for the constituents because, well,
+    ! they are constituents.
```
nusbaume: I might remove this comment, because hopefully some day in the future a scientist is only using CAM-SIMA and thus will have no idea what this comment is refering to.

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,356 @@
...
+    use tracer_data,               only: advance_trcdata
+
+    ! host model dependency for history output
+    use cam_history,               only: history_out_field
+
+    ! framework dependency for const_props
+    use ccpp_constituent_prop_mod, only: ccpp_constituent_prop_ptr_t
+
+    ! dependency to get constituent index
+    use ccpp_const_utils,          only: ccpp_const_get_idx
```
nusbaume: Given that this code is not going back to CAM, can we go ahead and use the CCPP-framework provided method instead?


```suggestion
    use ccpp_scheme_utils,          only: ccpp_constituent_index
```

Also if you make this change then I believe you no longer need to bring in the `const_props` variable into this subroutine as well.

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,356 @@
...
+    ! dependency for unit string handling
+    use string_utils,              only: to_lower, get_last_significant_char
+
+    integer,            intent(in)    :: ncol
+    integer,            intent(in)    :: pver
+    integer,            intent(in)    :: pcnst
+    type(ccpp_constituent_prop_ptr_t), &
+                        intent(in)    :: const_props(:)
+    real(kind_phys),    intent(in)    :: mwdry             ! molecular weight of dry air [g mol-1]
+    real(kind_phys),    intent(in)    :: boltz             ! Boltzmann constant [J K-1 molecule-1]
```
nusbaume: I would update the comment here to remove the reference to `molecule-1` (see explanation in later comment below):


```suggestion
    real(kind_phys),    intent(in)    :: boltz             ! Boltzmann constant [J K-1]
```

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,356 @@
...
+                         pmid, pint, phis, zi)
+
+    ! Get constituent indices for MMR and radius
+    call ccpp_const_get_idx(const_props, volcaero_const_name, &
+         mmr_idx, errmsg, errflg)
+    if (errflg /= 0) return
+
+    call ccpp_const_get_idx(const_props, volcrad_const_name, &
+         rad_idx, errmsg, errflg)
+    if (errflg /= 0) return
```
nusbaume: If you use the CCPP-framework provided subroutine then I believe these become:


```suggestion
    call ccpp_constituent_index(volcaero_const_name, &
         mmr_idx, errmsg=errmsg, errcode=errflg)
    if (errflg /= 0) return

    call ccpp_constituent_index(volcrad_const_name, &
         rad_idx, errmsg=errmsg, errcode=errflg)
    if (errflg /= 0) return
```

schemes/chemistry/prescribed_volcanic_aerosol.meta
```diff
@@ -0,0 +1,239 @@
...
+  dimensions = ()
+  intent = in
+[ prescribed_volcaero_file ]
+  standard_name = filename_for_prescribed_volcanic_aerosol
+  units = none
+  type = character | kind = len=*
+  dimensions = ()
+  intent = in
+[ volcaero_constituents ]
+  # or can this just be the ccpp_constituent_properties?
```
nusbaume: I don't believe you can directly use the `ccpp_constituent_properties` variable in the register phase as the variable hasn't been constructed yet, although @peverwhee would know for sure.

schemes/chemistry/prescribed_ozone.meta
nusbaume: I realize this will come off as somewhat esoteric, but I actually think the correct units for the Boltzmann Constant really is just `J K-1`, even though comments in CAM-SIMA and CESM_share says otherwise.  This is because the constant holds for any "particle" that can behave like a gas, which is technically not just molecules.  It's the same reason a number density is technically just a per-volume unit (e.g. `cm-3`), and not molecules per volume (e.g. `molecules cm-3`).

 So I might vote to revert this change, and eventually update the comment in `physconst.F90` in CAM-SIMA as well (we can probably leave CESM_share as-is for now, or I can update it in my eventual isotope physics PR).  Thanks!

nusbaume (changes requested):
Thanks for bringing this in @jimmielin!  I have some change requests but hopefully nothing major, and of course if you have any questions or concerns with any of my requests then just let me know.  Thanks again!

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thank you, removed!

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: I think this will matter in MAM too where the geometric-mean diameter of the climate list modes ("DGNUM") is stored as a pbuf field and I built in the assumption that it will also be a constituent in the modal aerosol state port in the bigger BAM/MAM optics PR. I agree it is best to have it as an attached property to each bin, and this slots well with the development of a host-side properties object. I will open an issue for this. Thanks for the heads up!


I think this is a bigger host-side issue so I opened an issue in CAM-SIMA: https://github.com/ESCOMP/CAM-SIMA/issues/496

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thanks! I waffled a bit on whether or not to remove this since I ran into a runtime crash if I registered history fields for the constituents here, but I guess it's a developer lesson. Removed!

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thanks, updated!

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thanks, removed!

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thanks, updated as suggested. Just to confirm it will be established convention to use keyword arguments for `errmsg`, `errcode` here or it would be OK to just write `errmsg, errflg)` as well?

schemes/chemistry/prescribed_volcanic_aerosol.meta
jimmielin: Thanks, I think this is a remnant paste from the other prescribed schemes when I was figuring out how to register dynamic constituents. I removed it here, and will remove the comments in the ozone/aerosol schemes in a later PR to avoid polluting the diff for this one.

schemes/chemistry/prescribed_ozone.meta
jimmielin: Thanks, removed! I can update the `physconst.F90` comment in CAM-SIMA in the companion PR https://github.com/ESCOMP/CAM-SIMA/pull/477

schemes/chemistry/prescribed_volcanic_aerosol.meta
nusbaume: Doesn't this entry need to be removed as well now that `const_props` is no longer in the call list?

schemes/chemistry/prescribed_volcanic_aerosol.F90
nusbaume: Good question!  I decided to use keyword arguments here because A.  They are both optional arguments, and using keywords helps indicate that, and B.  Their order is actually flipped, with `errflg` listed first in the actual subroutine call list.

However, I could easily be convinced to just use the regular argument method, it just seemed like the simplest way to "suggest" the code changes on Github in this particular situation.

nusbaume (approved):
Thanks for implementing my requests @jimmielin!  I believe you'll need to remove `const_props` from the metadata file, but otherwise everything looks good to me now.  Thanks again!

schemes/chemistry/prescribed_volcanic_aerosol.meta
jimmielin: Oops you're right, removed, thanks!

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thank you for explaining!

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,347 @@
...
+    character(len=*), parameter :: subname = 'prescribed_volcanic_aerosol_register'
+
+    errmsg = ''
+    errflg = 0
+
+    ! Check if prescribed volcanic aerosols are enabled
+    if (prescribed_volcaero_file == 'UNSET' .or. &
+        len_trim(prescribed_volcaero_file) == 0) then
+      if (amIRoot) then
+        write(iulog,*) subname//': No prescribed volcanic aerosols specified'
```
cacraigucar, line 79: Should errflg be set as well because it is currently `0`

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,347 @@
...
+    ! Register two constituents: aerosol MMR and geometric-mean radius
+    allocate(volcaero_constituents(2), stat=errflg, errmsg=errmsg)
+    if (errflg /= 0) then
+      errmsg = subname // ": " // trim(errmsg)
+      return
+    end if
+
+    ! (1) Volcanic aerosol dry mass mixing ratio
+    call volcaero_constituents(1)%instantiate( &
+         std_name          = volcaero_const_name, &
```
cacraigucar, line 95: Are these truly CCPP standard names?  If so, `VOLC_MMR` and `VOLC_RAD_GEOM` do not adhere to CCPP standard name structure.  On the other hand, that may not matter in this usage, I'm not sure.

schemes/chemistry/prescribed_volcanic_aerosol.F90
```diff
@@ -0,0 +1,347 @@
...
+         min_value         = 0.0_kind_phys, &
+         advected          = .false., &
+         water_species     = .false., &
+         mixing_ratio_type = 'dry', &
+         errcode           = errflg, &
+         errmsg            = errmsg)
+    if (errflg /= 0) return
+
+    if (amIRoot) then
+      write(iulog,*) trim(subname)//': Registered 2 prescribed volcanic aerosol constituents'
```
cacraigucar: Since they are hardwired, would it be good just to list the names of what has been registered instead of a generic `2 prescribed volcanic aerosol constituents`

schemes/chemistry/prescribed_volcanic_aerosol.meta
```diff
@@ -0,0 +1,232 @@
...
+  dimensions = ()
+  intent = in
+[ iulog ]
+  standard_name = log_output_unit
+  units = 1
+  type = integer
+  dimensions = ()
+  intent = in
+[ prescribed_volcaero_file ]
+  standard_name = filename_for_prescribed_volcanic_aerosol
```
cacraigucar: Current filename standard names in the git repo use `filename_of_...`.  `filename_for_...` might be appropriate, but just asking

cacraigucar (changes requested):

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thanks, this is intended to not error out - whether or not this module does anything is a namelist option; if the specifier is not set then `has_prescribed_volcaero` is kept as `.false.`.

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thanks for the question. This isn't a true standard name - I think we won't have proper standard names for chemistry/aerosol constituents, and keep just the same names as they are in CAM for now.

schemes/chemistry/prescribed_volcanic_aerosol.F90
jimmielin: Thanks, now writing the names into the log.

schemes/chemistry/prescribed_volcanic_aerosol.meta
jimmielin: Thanks for noticing! Changed to `filename_of_prescribed_volcanic_aerosol`

(https://github.com/ESCOMP/ESMStandardNames/blob/main/StandardNamesRules.rst specifies `filename_of_` so I will go with that)

cacraigucar (approved):

PR #397, opened 2026-05-18
Add MYNN PBL scheme to convection-permitting physics suite

kuanchihwang:
### Tag name

atmos_phys0_25_000

### Originator(s)

kuanchihwang

### Description (include the issue title, and the keyword ['closes', 'fixes', 'resolves'] followed by the issue number)

This PR adds MYNN PBL scheme to convection-permitting physics suite.

Additionally, an issue with the `sima_state_diagnostics` scheme has been identified and fixed. Consider the following two constituents for example:

* mass_number_concentration_of_cloud_liquid_water_particles_in_air
* mass_number_concentration_of_cloud_ice_water_crystals_in_air

The `sima_state_diagnostics` scheme adds constituents to history output by truncating their standard names to 32 characters. However, by doing so, both constituents will appear to have the same diagnostic name, "mass_number_concentration_of_clo". This issue will cause the model to crash.

Use the proper [`diagnostic_name`](https://github.com/NCAR/ccpp-framework/blob/sima_2026-04-13/src/ccpp_constituent_prop_mod.F90#L2019) API from the constituent object instead to fix it.

### List all namelist files that have been modified

None

### List all files eliminated and why

None

### List all files added and what they do

```
A       schemes/mmm/bl_mynn_compat.F90
A       schemes/mmm/bl_mynn_compat.meta
  * Add MYNN PBL scheme
```

### List all existing files that have been modified, and describe the changes

```
M       schemes/mmm/bl_gwdo_compat.F90
M       schemes/mmm/bl_gwdo_compat.meta
  * Consolidate constant initialization to `init` phase of YSU OGWD scheme
  * Keep consistent code style in YSU OGWD scheme
M       schemes/mmm/ccpp_kind_types.F90
  * Accommodate `kind_phys8` usage in MYNN PBL scheme
M       schemes/mmm/cu_ntiedtke_compat.F90
M       schemes/mmm/cu_ntiedtke_compat.meta
  * Remove `evap` from nTiedtke convection scheme
M       schemes/mmm/mmm_physics_compat.F90
M       schemes/mmm/mmm_physics_compat.meta
  * Add and modify interstitial schemes to support MYNN PBL scheme
M       schemes/mmm/sf_mynn_compat.F90
M       schemes/mmm/sf_mynn_compat.meta
  * Remove unnecessary array copying in MYNN surface layer scheme
  * Consolidate constant initialization to `init` phase of MYNN surface layer scheme
  * Change `spp_pbl` to integer in MYNN surface layer scheme
M       schemes/sima_diagnostics/sima_state_diagnostics.F90
  * Call proper API to get constituent diagnostic name
M       schemes/utilities/state_converters.F90
M       schemes/utilities/state_converters.meta
  * Add state converters between dry and wet snow mixing ratios
M       test/test_suites/suite_convection_permitting.xml
  * Add MYNN PBL scheme to convection-permitting physics suite
M       test/unit-test/tests/mmm/mmm_physics_compat_tests.pf
  * Fix unit tests
```

### List all automated tests that failed, as well as an explanation for why they were not fixed

None

### Is this an answer-changing PR? If so, is it a new physics package, algorithm change, tuning change, etc?

Yes, but only for convection-permitting physics suite.

### If yes to the above question, describe how this code was validated with the new/modified features

Configure, build, and run CAM-SIMA with the `MPASCP` compset.


schemes/mmm/sf_mynn_compat.meta
```diff
@@ -12,9 +12,9 @@
   dimensions = ()
   intent = in
 [ spp_pbl ]
-  standard_name = flag_for_stochastically_perturbed_parameterization_for_mynn_scheme
-  units = flag
-  type = logical
+  standard_name = control_for_stochastically_perturbed_parameterization_in_mynn_scheme
```
nusbaume, line 15: Just double-checking that this change is due to there being more than one option for `spp_pbl` in MMM physics?

schemes/mmm/mmm_physics_compat.F90
```diff
@@ -10,33 +10,46 @@ module mmm_physics_compat
...
         real(kind_phys), intent(out) :: xice_threshold
         character(*), intent(out) :: errmsg
         integer, intent(out) :: errflg
 
         errmsg = ''
         errflg = 0
 
-        ! These options are hardcoded to the same values as in the MPAS physics driver.
-        ! There are other possible values for them in WRF, but the following combination is the only supported one in MPAS.
+        scheme_name = 'mmm_physics_compat'
```
nusbaume: I actually think `scheme_name` does need to be in the `_run` phase subroutine.  The reason is because there are other physics schemes which output a `scheme_name` variable during the run phase, which means this variable would immediately get clobbered by a different scheme if scientists did any future mixing-and-matching of schemes in a suite.

schemes/mmm/mmm_physics_compat.F90
```diff
@@ -258,6 +313,57 @@ pure subroutine compute_characteristic_grid_length_scale_init( &
         dx(:) = sqrt(omega(:) * (rearth ** 2))
     end subroutine compute_characteristic_grid_length_scale_init
 
+    !> \section arg_table_compute_hydrostatic_upward_air_velocity_at_if_run Argument Table
+    !! \htmlinclude compute_hydrostatic_upward_air_velocity_at_if_run.html
+    pure subroutine compute_hydrostatic_upward_air_velocity_at_if_run( &
```
nusbaume: Nit-pick, but can we change `if` to something like `intf`?  When I first read the name of the subroutine I thought it was some sort of conditional statement.  Thanks!

schemes/mmm/mmm_physics_compat.meta
```diff
@@ -5,6 +5,24 @@
 [ccpp-arg-table]
   name = mmm_physics_compat_init
   type = scheme
+[ initial_run ]
+  standard_name = is_initial
```
nusbaume: I might add "run" here to make it a little more specific, along with an extra long name so we can make sure CESM users understand what it is as well:

 
```suggestion
  standard_name = is_initial_run
  long_name = Flag indicating whether this simulation is an initial (startup) run, or a continuation of a previous run.
```

schemes/mmm/mmm_physics_compat.meta
```diff
@@ -24,7 +42,19 @@
...
+  dimensions = ()
+  intent = out
+[ cycling ]
+  standard_name = is_data_assimilation_cycling_for_mmm_scheme
+  units = flag
+  type = logical
+  dimensions = ()
+  intent = out
+[ restart ]
+  standard_name = is_restart_for_mmm_scheme
```
nusbaume: I might add "run" here as well:


```suggestion
  standard_name = is_restart_run_for_mmm_scheme
```

schemes/mmm/mmm_physics_compat.meta
```diff
@@ -233,18 +257,72 @@
...
   dimensions = (horizontal_dimension, vertical_layer_dimension)
   intent = out
+[ rqiblten ]
+  standard_name = tendency_of_cloud_ice_mixing_ratio_wrt_dry_air_due_to_pbl_processes
+  units = kg kg-1 s-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  intent = out
+[ rniblten ]
+  standard_name = tendency_of_mass_number_concentration_of_cloud_ice_water_crystals_in_air_due_to_pbl_processes
```
nusbaume: Just double checking if this is with respect to total air or dry air?  If it is dry air then we should update the standard name like so:

 
```suggestion
  standard_name = tendency_of_mass_number_concentration_of_cloud_ice_water_crystals_in_dry_air_due_to_pbl_processes
```

schemes/mmm/mmm_physics_compat.meta
```diff
@@ -233,18 +257,72 @@
...
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  intent = out
+[ rozblten ]
+  standard_name = tendency_of_ozone_mixing_ratio_wrt_moist_air_and_condensed_water_due_to_pbl_processes
+  units = kg kg-1 s-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  intent = out
+[ rnwfablten ]
+  standard_name = tendency_of_mass_number_concentration_of_hygroscopic_aerosols_due_to_pbl_processes
```
nusbaume: Just double-checking that these aerosol number concentrations are with respect to total air, and not dry air?   If they are with respect to dry air then we should update the standard names to indicate that.

schemes/mmm/mmm_physics_compat.meta
```diff
@@ -233,18 +257,72 @@
...
   dimensions = (horizontal_dimension, vertical_layer_dimension)
   intent = out
+[ rqcblten ]
+  standard_name = tendency_of_cloud_liquid_water_mixing_ratio_wrt_dry_air_due_to_pbl_processes
+  units = kg kg-1 s-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  intent = out
+[ rncblten ]
+  standard_name = tendency_of_mass_number_concentration_of_cloud_liquid_water_particles_in_air_due_to_pbl_processes
```
nusbaume: I realize this matches what is in the official dictionary, but I just wanted to double-check if these number concentrations are with respect to all air, or just dry air? If it is just dry air then I think we'll need to update the standard name like so:


```suggestion
  standard_name = tendency_of_mass_number_concentration_of_cloud_liquid_water_particles_in_dry_air_due_to_pbl_processes
```

schemes/mmm/mmm_physics_compat.meta
```diff
@@ -317,6 +395,49 @@
...
+  intent = inout
+[ ozone ]
+  standard_name = ozone_mixing_ratio_wrt_moist_air_and_condensed_water
+  units = kg kg-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = inout
+  advected = True
+[ nc ]
+  standard_name = mass_number_concentration_of_cloud_liquid_water_particles_in_air
```
nusbaume: Just checking one last time that all of the mass number concentration variables here and below are with respect to total air, and not dry air?

test/test_suites/suite_convection_permitting.xml
```diff
@@ -21,24 +24,32 @@
     <scheme>sf_mynn_compat</scheme>
     <scheme>sf_mynn_diagnostics</scheme>
 
+    <!-- MYNN PBL scheme -->
+    <scheme>bl_mynn_compat_pre</scheme>
+    <scheme>bl_mynn_compat</scheme>
+    <scheme>bl_mynn_diagnostics</scheme>
```
nusbaume, line 31: I sincerely apologize for not catching this in previous PRs, but the MMM diagnostics schemes (e.g. `bl_mynn_diagnostics` here) needs to be moved into their own CCPP-ized Fortran module file(s). 

The reason is because the subroutines `history_add_field` and `history_out_field` are SIMA-specific subroutines that are not (currently) planned to be part of the larger CCPP ecosystem.  Thus by moving them into a separate module we can keep the core physics schemes portable across host models.  

For CAM physics we have been putting the diagnostics modules under `schemes/sima_diagnostics`, but I am happy to have the MMM diagnostics live wherever you think is most appropriate.

Anyways, I realize this might take a lot of work, so I'm happy to put this in a "future work" issue and deal with it later.  I just wanted to record it here before I forgot about it.

schemes/mmm/bl_mynn_compat.F90
```diff
@@ -0,0 +1,784 @@
+!> This module contains interstitial schemes that are specific to MYNN PBL scheme,
```
nusbaume: Grammar typo:


```suggestion
!> This module contains interstitial schemes that are specific to the MYNN PBL scheme,
```

schemes/mmm/bl_mynn_compat.F90
```diff
@@ -0,0 +1,784 @@
...
+!> which is part of MMM physics.
+module bl_mynn_compat
+    use ccpp_kinds, only: kind_phys
```
nusbaume: Given that there are no module-level variables that need `kind_phys`, can we move this `use` statement down to the relevant subroutine level instead?

schemes/mmm/bl_mynn_compat.F90
```diff
@@ -0,0 +1,784 @@
...
+
+    implicit none
+
+    private
+    public :: bl_mynn_compat_pre_run
+    public :: bl_mynn_compat_init
+    public :: bl_mynn_compat_run
+    public :: bl_mynn_diagnostics_init
+    public :: bl_mynn_diagnostics_run
```
nusbaume: Just adding another note here that we should try and move these SIMA-specific diagnostics schemes into a separate module, to help with the portability of the core physics routines (although I am happy if we want to do this move in a future PR).

schemes/mmm/bl_mynn_compat.F90
```diff
@@ -0,0 +1,784 @@
...
+
+        flag_qc = .false.
+        flag_qi = .false.
+        flag_qs = .false.
+        flag_qoz = .false.
+
+        call ccpp_constituent_index( &
+            'cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water', i, errflg, errmsg)
+
+        if (errflg == 0) then
```
nusbaume: Instead of checking `errflg`, I think you'll instead want to check if the returned index is different from `int_unassigned`:


```suggestion
        use ccpp_constituent_prop_mod, only: int_unassigned
        ...
        if (i /= int_unassigned) then
```

I would make this same modification for the other index checks below as well.

schemes/mmm/bl_mynn_compat.F90
```diff
@@ -0,0 +1,784 @@
...
+        real(kind_phys), intent(out) :: exch_h(:, :), exch_m(:, :), &
+                                        dqke(:, :), qwt(:, :), qshear(:, :), &
+                                        qbuoy(:, :), qdiss(:, :), &
+                                        maxwidth(:), &
+                                        maxmf(:), ztop_plume(:)
+        character(*), intent(out) :: errmsg
+        integer, intent(out) :: errflg
+
+        integer, parameter :: kts = 1
+        real(kind_phys), allocatable :: sqv(:, :), sqc(:, :), sqi(:, :), sqs(:, :)
```
nusbaume: Instead of making these `allocatable`, I think it might be better to just make them local stack variables instead:

```suggestion
        real(kind_phys) :: sqv(size(sqv_dry,1), size(sqv_dry,2)), sqc(size(sqc_dry,1), size(sqc_dry,2)), sqi(size(sqi_dry,1), size(sqi_dry,2)), sqs(size(sqs_dry,1), size(sqs_dry,2))
```

This will then allow you to remove the `allocate` and error-checking code below (while arguably maintaining the same memory overhead and slightly improving performance).

schemes/mmm/bl_mynn_compat.meta
```diff
@@ -0,0 +1,1571 @@
+[ccpp-table-properties]
+  name = bl_mynn_compat_pre
+  type = scheme
+  dependencies = ccpp_kind_types.F90, mmm_physics/mynn_shared.F90, mmm_physics/bl_mynn_subroutines.F90, mmm_physics/bl_mynn.F90
```
nusbaume: I don't _think_ this `dependencies` line is needed here, as the `bl_mynn_compat_pre_run` scheme doesn't use any of these dependencies.

schemes/mmm/bl_mynn_compat.meta
```diff
@@ -0,0 +1,1571 @@
...
+  intent = in
+[ initflag ]
+  standard_name = control_for_initialization_in_mynn_pbl_scheme
+  units = 1
+  type = integer
+  dimensions = ()
+  intent = out
+[ kpbl ]
+  standard_name = vertical_index_at_top_of_atmosphere_boundary_layer
+  units = count
```
nusbaume: I believe the official units for all index variables is `index`:


```suggestion
  units = index
```

This same request holds for the other "vertical_index" variable below as well.

schemes/mmm/bl_mynn_compat.meta
```diff
@@ -0,0 +1,1571 @@
...
+  dimensions = ()
+  intent = out
+[ bl_mynn_stfunc ]
+  standard_name = control_for_stability_function_in_mynn_pbl_scheme
+  units = 1
+  type = integer
+  dimensions = ()
+  intent = out
+[ bl_mynn_tkeadvect ]
+  standard_name = flag_for_tke_advection_in_mynn_pbl_scheme
```
nusbaume: Can we turn the `flag_for` standard names here and below into verbs, like `do_tke_advection_in_mynn_pbl_scheme`?  That way it is really clear what "True" and "False" mean for that particular logical.

schemes/mmm/bl_mynn_compat.meta
```diff
@@ -0,0 +1,1571 @@
...
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  intent = out
+[ sm ]
+  standard_name = stability_function_for_momentum
+  units = 1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_dimension, vertical_layer_dimension)
+  intent = out
+[ qc_bl ]
+  standard_name = subgrid_scale_cloud_liquid_water_mixing_ratio_wrt_moist_air
```
nusbaume, line 502: Can you describe how these "subgrid_scale" variables are different from the standard mixing ratio variables?  I am mostly just asking because they appear to have the same dimensions as the regular mixing ratio variables.  Thanks!

test/test_suites/suite_convection_permitting.xml
```diff
@@ -7,12 +7,15 @@
     <scheme>calc_exner</scheme>
     <scheme>calc_hydrostatic_air_density</scheme>
```
nusbaume: I realize this is not directly related to this PR, but there was some discussion that the air density coming from the hydrostatic equation may not always be valid at high resolutions when the atmosphere is sometimes non-hydrostatic.  Given this, would it be better to use the ideal gas law, or whatever method MPAS-A is using to calculate density instead?

Note that this doesn't have to be done in this PR, as it certainly isn't critical (so we can just open an issue instead if preferred).   Thanks!

schemes/mmm/bl_mynn_compat.meta
```diff
@@ -0,0 +1,1571 @@
...
+  intent = inout
+[ maxwidth ]
+  standard_name = maximum_plume_width_for_mynn_pbl_scheme
+  units = m
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent)
+  intent = out
+[ maxmf ]
+  standard_name = maximum_mass_flux_for_mynn_pbl_scheme
+  units = m s-1
```
nusbaume, line 1175: I generally wouldn't assume that mass flux has units of velocity (`m s-1`).  Are the units correct, and if so, is an assumption being made here to remove the presence of mass in the units (i.e. `kg`)?  

If there is an additional assumption then it might be good to just add a `long_name` attribute describing it.

nusbaume (changes requested):
Thanks for bringing in MYNN @kuanchihwang!  I have some change requests that are somewhat major, but they can certainly be pushed off onto future work.  Otherwise everything else shouldn't be too difficult (but of course please let me know if that is not the case).  Thanks again!

schemes/mmm/mmm_physics_compat.F90
kuanchihwang: I moved `scheme_name` initialization back to run phase in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/373139f0232ea3c0e24c1e487aef963d2a72ce5a.

schemes/mmm/mmm_physics_compat.F90
kuanchihwang: I added `at_interface` to the subroutine names in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/19b929118a8219f4c81bee12f3d396c444c180e1.

schemes/mmm/mmm_physics_compat.meta
kuanchihwang: I added "_run" to those standard names in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/66b6c4dda9c67f6796f3ddc1f476a866b98b7624.

schemes/mmm/mmm_physics_compat.meta
kuanchihwang: As above, see https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/66b6c4dda9c67f6796f3ddc1f476a866b98b7624. However, [`is_restart`](https://github.com/ESCOMP/ESMStandardNames/blob/v1.0.0/Metadata-standard-names.md?plain=1#L1187-L1188) is ratified in the ESM Standard Names.

test/test_suites/suite_convection_permitting.xml
kuanchihwang: I split out the diagnostic schemes in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/0aea11c9fe3d278ad925927c3c974345a337cd32.

schemes/mmm/bl_mynn_compat.F90
kuanchihwang: I fixed the grammar in header code comments here and elsewhere in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/bb2329e5ed340ae85f3c029facf5d742b70989db.

schemes/mmm/bl_mynn_compat.F90
kuanchihwang: I removed use statements at module level in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/2d5848aedb252c32a385568fea2b00bb2c133ad3.

schemes/mmm/bl_mynn_compat.F90
kuanchihwang: As above, I split out the diagnostic schemes in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/0aea11c9fe3d278ad925927c3c974345a337cd32.

schemes/mmm/bl_mynn_compat.meta
kuanchihwang: Indeed, this dependency line is not needed. I removed it in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/43d802d20641574d1ae84b7714e909c630fee492.

schemes/mmm/bl_mynn_compat.F90
kuanchihwang: I added additional checks against `int_unassigned` in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/05780bf99279aa34d4b86cdac6d5ef93eaaf4177.

schemes/mmm/bl_mynn_compat.meta
kuanchihwang: I changed the units of index variables to "index" in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/ce59f7cf60c3f2aa8ef0465bdb1891b7f99f203c.

schemes/mmm/bl_mynn_compat.meta
kuanchihwang: Sure, I changed the standard names to use the "do_" prefix in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/5a9d5b4451edc4f3feb1d91a7e8c9b0fa7a41a75.

Is this naming convention documented anywhere? The ESM Standard Names suggests that:

> For control-oriented variables, if the variable is a Fortran logical, use "flag_for_X". If it is any other data type, use "control_for_X". All flags should be Fortran logicals.

And, "do_" is not on the list of known prefixes.

test/test_suites/suite_convection_permitting.xml
kuanchihwang: I implemented the conversion schemes between temperature and virtual temperature in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/0e7dd4a7bb72ae4b30907672a98303516801a479. The implementation is verified by [unit tests](https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/d8d95613c22540d06940431f883f62db2fa74a61).

Therefore, we can now calculate full air density directly by equation of state in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/44e1d27b1137420421e3566422ae9df54f3910ab. Accordingly, the `calc_hydrostatic_air_density` scheme was removed.

schemes/mmm/bl_mynn_compat.meta
kuanchihwang: Yes, many variables in the MYNN SFC/PBL schemes are definitely counterintuitive...

Code-wise, `maxmf` is calculated as `edmf_a * edmf_w` in the [`dmp_mf`](https://github.com/NCAR/MMM-physics/blob/20251208-CAM-SIMA/bl_mynn_subroutines.F90#L5484-L5485) subroutine. `edmf_a` has units of (area) "fraction", and `edmf_w` has units of "m s-1".

Therefore, we can multiply `maxmf` by the bottom area of a column to get volumetric flux (m3 s-1). Then, multiply volumetric flux by air density to arrive at mass flux (kg s-1).

schemes/mmm/sf_mynn_compat.meta
kuanchihwang: I changed the `spp_pbl` variable to integer to more closely follow what MPAS does. In the MPAS physics driver, `spp_pbl` is declared as an integer control variable<sup>[[1]](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/physics/mpas_atmphys_vars.F#L417)</sup>. But unfortunately, the MYNN SFC/PBL schemes use it inconsistently.

The MYNN SFC scheme inexplicably expects it to be logical<sup>[[2]](https://github.com/NCAR/MMM-physics/blob/20251208-CAM-SIMA/sf_mynn.F90#L164)</sup>, while the MYNN PBL scheme expects it to be integer<sup>[[3]](https://github.com/NCAR/MMM-physics/blob/20251208-CAM-SIMA/bl_mynn.F90#L192-L193)</sup>. Therefore, I opted to account for this inconsistency directly in the `sf_mynn_compat` compatibility layer<sup>[[4]](https://github.com/ESCOMP/atmospheric_physics/blob/635560d2436fc0273691fb486d94933c66a08052/schemes/mmm/sf_mynn_compat.F90#L223)</sup> like in MPAS<sup>[[5]](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.0/src/core_atmosphere/physics/physics_wrf/module_sf_mynn.F#L231-L232)</sup>.

schemes/mmm/bl_mynn_compat.meta
dudhia: These variables are only seen by the radiation and not the microphysics.

schemes/mmm/bl_mynn_compat.F90
kuanchihwang: These arrays are three-dimensional, and their size can become very large at high-resolution simulations. Using big automatic arrays in Fortran can be problematic because different compilers implement the memory allocation behavior differently.

For example, the Intel Fortran compiler puts automatic arrays on the stack by default. When the size of automatic arrays exceeds the stack limit, a segmentation fault will occur. Most Linux distributions have a default stack limit of 8192 KB, which is easily exceeded if we use automatic arrays this way.

Below is a proof of concept demonstrating this issue:

```bash
$ cat auto_arrays.f90
program auto_arrays
    implicit none

    integer, parameter :: n = 8 * 1024 * 1024 / 4

    call test_auto_arrays(n)
contains
    subroutine test_auto_arrays(n)
        integer, intent(in) :: n
        integer :: a(n)

        print *, size(a)
    end subroutine test_auto_arrays
end program auto_arrays
$ ulimit -s
8192
$ ifx --version
ifx (IFX) 2025.3.3 20260319
Copyright (C) 1985-2026 Intel Corporation. All rights reserved.
$ ifx -g -O0 -traceback auto_arrays.f90
$ ./a.out
forrtl: severe (174): SIGSEGV, segmentation fault occurred
Image              PC                Routine            Line        Source
libc.so.6          00007D6A5B1B4C30  Unknown               Unknown  Unknown
a.out              000000000040530F  test_auto_arrays           12  auto_arrays.f90
a.out              0000000000405266  auto_arrays                 6  auto_arrays.f90
a.out              000000000040522D  Unknown               Unknown  Unknown
libc.so.6          00007D6A5B19F610  Unknown               Unknown  Unknown
libc.so.6          00007D6A5B19F6C0  __libc_start_main     Unknown  Unknown
a.out              0000000000405145  Unknown               Unknown  Unknown
$ ulimit -s unlimited
$ ./a.out
     2097152
```

The same code snippet does not crash when compiled by the GNU Fortran compiler because it puts automatic arrays *of unknown sizes* on the heap instead.

Therefore, I would be wary about switching to automatic arrays here due to the inconsistent compiler behavior.

schemes/mmm/mmm_physics_compat.meta
dudhia: Correct to use dry air here as MPAS always is wrt dry air. Some ambiguity in MYNN as it does not convert the number variables to per moist air while it does for mass variables.

schemes/mmm/mmm_physics_compat.meta
dudhia: yes, as above, we want these to be per dry air in MPAS

schemes/mmm/mmm_physics_compat.meta
dudhia: yes, as above, number should be per dry air.

schemes/mmm/mmm_physics_compat.meta
dudhia: yes, per dry air mass

schemes/mmm/mmm_physics_compat.meta
kuanchihwang: I updated the standard names in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/0a5dd84be3ecd2becea05da416a2083de1026a85.

schemes/mmm/mmm_physics_compat.meta
kuanchihwang: As above, see https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/0a5dd84be3ecd2becea05da416a2083de1026a85.

schemes/mmm/mmm_physics_compat.meta
kuanchihwang: As above, see https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/0a5dd84be3ecd2becea05da416a2083de1026a85.

schemes/mmm/mmm_physics_compat.meta
kuanchihwang: As above, see https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/0a5dd84be3ecd2becea05da416a2083de1026a85.

schemes/mmm/bl_mynn_compat.F90
nusbaume: While that is a totally fair point, as you show in your example one can just increase the allowed stack size and have it run without an issue.  Plus there are certain CAM schemes which have many (>100) 3-D stack variables, so if the stack size limit becomes a problem we'll have issues well beyond the MYNN scheme.  On top of that allocating and deallocating every physics timestep, especially for large arrays and short time steps that might show up in high-resolution runs, is going to be a non-trivial performance hit.

So I think I am still leaning towards making them stack variables that may cause a problem in certain environments or machines, then leaving them as heap variables that will impact performance regardless of the environment.

schemes/mmm/bl_mynn_compat.meta
nusbaume: Thanks for the changes, and for pointing out the standard name rules inconsistency!  I'll bring it up in the next standard names meeting to see about getting the official rules updated.

schemes/mmm/bl_mynn_compat.meta
nusbaume: Thanks for the explanation @dudhia!  The variable naming seems reasonable to me then.

schemes/utilities/state_converters.F90
```diff
@@ -10,11 +10,15 @@ module state_converters
...
+  public :: temp_to_virtual_temp_run
+  public :: virtual_temp_to_temp_run
+
   ! Calculate dry air density by equation of state/ideal gas law
   public :: calc_dry_air_ideal_gas_density_run
 
-  ! Calculate air density by hydrostatic equation
-  public :: calc_hydrostatic_air_density_run
+  ! Calculate air density by equation of state/ideal gas law
+  public :: calc_air_density_run
```
nusbaume: Can we rename this to `calc_air_ideal_gas_density_run` to make it even clearer as to how the density is being  calculated?

schemes/mmm/bl_mynn_compat.meta
nusbaume: Thanks for the explanation!  I'll leave this comment as-is to help as a future reference, but will consider it "resolved" for the purpose of this PR.

nusbaume (approved):
Thanks for the updates and new air density schemes @kuanchihwang!  I had a couple minor requests/comments, but nothing that should block this PR from being merged.  Thanks again!

schemes/utilities/state_converters.F90
kuanchihwang: Yes, I renamed it in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/f2357ae2fb8c27909fc501e78c1a70d4942e7650.

schemes/mmm/bl_mynn_compat.F90
kuanchihwang: I switched them to automatic arrays in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/14769fa38972135ee52e57c55d4a58283bc39931.

schemes/mmm/bl_mynn_compat.meta
```diff
@@ -0,0 +1,1300 @@
...
+  units = index
+  type = integer
+  dimensions = (horizontal_loop_extent)
+  intent = out
+[ ktop_plume ]
+  standard_name = vertical_index_at_top_of_tallest_plume_for_mynn_pbl_scheme
+  units = index
+  type = integer
+  dimensions = (horizontal_loop_extent)
+  intent = out
```
jimmielin, line 55: Can I ask if `kpbl` and `ktop_plume` here are in which vertical coordinate order (is it 1 as TOA or 1 as lowest layer)?

The reason I ask is that they appear to be out directly from `bl_mynn_run`, so if `bl_mynn_run` outputs as 1 as lowest layer, then this `vertical_index_` would be inverted compared to what is conventionally indexed in CAM-SIMA.

jimmielin (commented):
Thanks @kuanchihwang! I just had one question before I approve the PR.

schemes/mmm/bl_mynn_compat.meta
kuanchihwang: Nice catch! Thanks for spotting it. I inverted the vertical indexes in https://github.com/ESCOMP/atmospheric_physics/pull/397/commits/bbef3280d7e5b194a8c8d612dce17b071295e09f.

jimmielin (approved):
Thanks @kuanchihwang!

PR #399, opened 2026-05-27
Prevent a segfault in ccpp tuvx wrapper

K20shores:
Originator(s): Kyle SHores

In testing the terminator configuration of musica, I found with gcc 14.3.0, that the tuvx ccpp wrapper code would segfault. This is because the result of the `size(...)` call, and the math around it, was coming back either as zero or a negative number ( I can't remember which now). This results in a segfault. I guess this was the result of an overzealous optimizer path, or maybe some fortran language specification that I'm unaware of.

Either way, saving the math result to a local variable resolved the segfault.

schemes/musica/tuvx/musica_ccpp_tuvx.F90


jimmielin (approved):
Looks like a odd compiler bug. Thanks for fixing this.

cacraigucar (approved):

PR #411, opened 2026-06-11
atmos_phys0_26_000: Move ndrop_bam, nucleate_ice, and other microp_aero interstitials to atmos_phys

jimmielin:
Originator(s): @jimmielin 

- Moves `nucleate_ice.F90` portable code to atmospheric_physics.
- Refactor and clean up `ndrop_bam.F90` to be portable and move to atmos_phys.
- CCPPize interstitial schemes used in `microp_aero`: `compute_subgrid_vertical_velocity`, `scale_subgrid_vertical_velocity`, `dust_default_radii`

```
A       schemes/microp_aero/scale_subgrid_vertical_velocity_namelist.xml
```

```
A       schemes/microp_aero/nucleate_ice.F90
  - portable nucleate_ice code

A       schemes/microp_aero/ndrop_bam.F90
  - portable ndrop_bam code

A       schemes/microp_aero/compute_subgrid_vertical_velocity.F90
A       schemes/microp_aero/compute_subgrid_vertical_velocity.meta
A       schemes/microp_aero/scale_subgrid_vertical_velocity.F90
A       schemes/microp_aero/scale_subgrid_vertical_velocity.meta
A       schemes/microp_aero/dust_default_radii.F90
A       schemes/microp_aero/dust_default_radii.meta
  - CCPPized interstitial schemes used in microp_aero

M       test/musica/test_musica_api.F90
M       test/musica/tuvx/test_tuvx_load_species.F90
M       to_be_ccppized/ccpp_const_utils.F90
  - fix blocking MUSICA tests breaking due to upstream CCPP framework update
```

All new files.

B4B in CAM
New chemistry/physics code (microp_aero)

If yes to the above question, describe how this code was validated with the new/modified features: CAM regression tests


schemes/microp_aero/scale_subgrid_vertical_velocity.F90
```diff
@@ -0,0 +1,68 @@
+! Apply min/max/scale to subgrid vertical velocity and derive wsubi.
```
fvitt: What is `wsubi`?  It might be helpful to elaborate here.

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,628 @@
...
+    sixth = 1._kind_phys/6._kind_phys
+    sq2 = sqrt(2._kind_phys)
+    pi = 4._kind_phys*atan(1.0_kind_phys)     ! note: this should probably use physconst pi but here for b4b
+    sqpi = sqrt(pi)
+    surften = 0.076_kind_phys
+    aten = 2._kind_phys*mwh2o*surften/(r_universal*tmelt*rhoh2o)
+    alogaten = log(aten)
+    alog2 = log(2._kind_phys)
+    alog3 = log(3._kind_phys)
+    super(:) = 0.01_kind_phys*supersat(:)
```
fvitt, line 163: Should these be declared and set as parameters at the module level?

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,628 @@
...
+        write (iulog, *) 'ndrop_bam_init: SULFATE aerosol properties FOUND at index ', idxsul
+      end if
+    end if
+
+    ! set parameters for droplet activation,
+    ! following abdul-razzak and ghan 2000, JGR
+    third = 1._kind_phys/3._kind_phys
+    sixth = 1._kind_phys/6._kind_phys
+    sq2 = sqrt(2._kind_phys)
+    pi = 4._kind_phys*atan(1.0_kind_phys)     ! note: this should probably use physconst pi but here for b4b
```
fvitt: Interesting that `physconst pi` is not bit-for-bit the same

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,591 @@
...
+!  With help from C. C. Chen and B. Eaton (2014)
+!-------------------------------------------------------------------------------
+
+use wv_saturation,  only: svp_water, svp_ice
+
+implicit none
+private
+save
+
+integer, parameter :: r8 = selected_real_kind(12)
```
fvitt: Is there a reason the define the real kind here and not use the `ccpp_kinds phys_kind`.  Do the reals here need to always be double precession?

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,591 @@
...
+real(r8) :: pi
+real(r8) :: mincld
+
+real(r8), parameter :: Shet   = 1.3_r8     ! het freezing threshold
+real(r8), parameter :: rhoice = 0.5e3_r8   ! kg/m3, Wpice is not sensitive to rhoice
+real(r8), parameter :: minweff= 0.001_r8   ! m/s
+real(r8), parameter :: gamma1=1.0_r8
+real(r8), parameter :: gamma2=1.0_r8
+real(r8), parameter :: gamma3=2.0_r8
+real(r8), parameter :: gamma4=6.0_r8
```
fvitt: It would be nice to document here what these parameters are.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,591 @@
...
+! parameters
+
+      A11 = 0.0263_r8
+      A12 = -0.0185_r8
+      A21 = 2.758_r8
+      A22 = 1.3221_r8
+      B11 = -0.008_r8
+      B12 = -0.0468_r8
+      B21 = -0.2667_r8
+      B22 = -1.4588_r8
```
fvitt: Should these be declared as parameters?   It would be nice to document where these numbers come from.  Maybe from the references at the top of the module?

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,591 @@
...
+      C1_fast  =0.0739_r8
+      C2_fast  =1.2372_r8
+
+      A1_slow  =-0.3949_r8
+      A2_slow  =1.282_r8
+      B1_slow  =-0.0156_r8
+      B2_slow  =0.0111_r8
+      B3_slow  =0.0217_r8
+      C1_slow  =0.120_r8
+      C2_slow  =2.312_r8
```
fvitt: `real(r8), parameter :: ...`

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,591 @@
...
+   DLOSS = FPIVOLc * C * B1 * R**2 / ( 1.0_r8+ B2 * R )
+   VICE  = ( A2 + A3 * S ) * DLOSS / ( A1 * S )  ! 2006,(19)
+   V_out = VICE*1e-2_r8  ! cm/s => m/s
+
+END SUBROUTINE Vpreice
+
+subroutine frachom(Tmean,RHimean,detaT,fhom)
+   ! How much fraction of cirrus might reach Shom
+   ! base on "A cirrus cloud scheme for general circulation models",
+   ! B. Karcher and U. Burkhardt 2008
```
fvitt: Include DOI if available.

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, this is pre-existing from CAM's copy: https://github.com/ESCOMP/CAM/blob/cf06897968961fee9b16201831c4d3bc1ba003a0/src/physics/cam/nucleate_ice.F90#L38

I think this file was not modified and was moved here directly

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks for the suggestion. I am worried that `sqrt()` and `atan()` and `log()` might be evaluated differently as parameter (during compile time) vs. at runtime leading to bit differences. Also some of it depends on the subroutine arguments (`mwh2o`, `r_universal`, ...) so we would be splitting the set of parameters into some `parameter` (third, sixth) and set at runtime (aten, sqpi, ...) so I would prefer to keep this block as-is to avoid accidental answer differences.

schemes/microp_aero/scale_subgrid_vertical_velocity.F90
jimmielin: Thanks, added comment

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks, this comment may be incorrect after actually testing `gfortran` and `ifx`. I flagged this for future cleanup instead: `! note: should use physconst pi in future cleanup.`

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, added notes

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, this is from Liu and Penner page 509 - added comment

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, made into `parameter`

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, added

schemes/microp_aero/ndrop_bam.F90
fvitt: Okay

schemes/microp_aero/nucleate_ice.F90
fvitt: Shouldn't this be updated to use `ccpp_kinds` for the real kind so that single precision could be used?

jimmielin: Done!

fvitt (approved):

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
```diff
@@ -0,0 +1,145 @@
...
+module compute_subgrid_vertical_velocity
+
+  use ccpp_kinds, only: kind_phys
+
+  implicit none
+  private
+
+  public :: compute_subgrid_vertical_velocity_tke_run
+  public :: compute_subgrid_vertical_velocity_kvh_run
+  public :: compute_subgrid_vertical_velocity_clubb_run
```
nusbaume, line 9: I believe all three of these subroutines can be labeled `pure`.

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
```diff
@@ -0,0 +1,145 @@
...
+    errflg = 0
+
+    wsub(:, :top_lev-1) = 0._kind_phys
+
+    do k = top_lev, pver
+      do i = 1, ncol
+        wsub(i,k) = sqrt(0.5_kind_phys * (tke(i,k) + tke(i,k+1)) * (2._kind_phys / 3._kind_phys))
+        wsub(i,k) = min(wsub(i,k), 10._kind_phys)
+      end do
+    end do
```
nusbaume, line 56: I might make a future-work issue noting that for both here and `compute_subgrid_vertical_velocity_clubb_run` they are using a standard equation that assumes turbulence is isotropic, and thus the vertical velocity is `sqrt(tke*2/3)`.  Given that function is a generally useful calculation, it might help to eventually make it a function and move it to `atmos_phys_pbl_utils.F90`.

Until then, I might just add a comment here noting that this calculation assumes the turbulent motion is isotropic, and that vertical velocity is limited to 10 m s-1.  Thanks!

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
```diff
@@ -0,0 +1,145 @@
...
+    ! Local variables
+    integer :: i, k
+
+    errmsg = ''
+    errflg = 0
+
+    wsub(:, :top_lev-1) = 0._kind_phys
+
+    ! Get sub-grid vertical velocity from diffusion coefficient.
+    ! Following Morrison et al. 2005, JAS.
```
nusbaume: Can we include the equation number and DOI here to help with provenance?


```suggestion
    ! Following equation 24 in Morrison et al. 2005, JAS,  https://doi.org/10.1175/JAS3446.1
```

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
```diff
@@ -0,0 +1,145 @@
...
+    tke(:ncol,1:top_lev-1) = 0._kind_phys
+
+    wsub(:, :top_lev-1) = 0._kind_phys
+
+    do k = top_lev, pver
+      do i = 1, ncol
+        wsub(i,k) = sqrt(0.5_kind_phys * (tke(i,k) + tke(i,k+1)) * (2._kind_phys / 3._kind_phys))
+        wsub(i,k) = min(wsub(i,k), 10._kind_phys)
+      end do
+    end do
```
nusbaume: Is this what CAM is currently doing now?  If so then we might need to talk with @adamrher or @bstephens82, as I think (?) CLUBB can now flip the vertical grid for CAM and CAM-SIMA, which means the flipping done in the first line here might introduce incorrect results.

On top of that, some of these calculations seem un-necessary, as they are essentially converting vertical velocity to TKE and then back to vertical velocity.  While I realize we need to keep this as-is to match CAM, it would be good to eliminate these extra calculations eventually.  Maybe worth adding to the "future work" issue?

Finally, I might add a comment here noting that these calculations assume the turbulent motion is isotropic.  Thanks again!

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
```diff
@@ -0,0 +1,145 @@
+! Compute raw subgrid-scale vertical velocity from TKE, KVH, or CLUBB WP2.
+module compute_subgrid_vertical_velocity
+
+  use ccpp_kinds, only: kind_phys
```
nusbaume: Missed this earlier, but I might move this `use` statement down to the subroutine level.

schemes/microp_aero/dust_default_radii.F90
```diff
@@ -0,0 +1,40 @@
...
+  real(kind_phys), parameter :: rn_dst1 = 0.258e-6_kind_phys
+  real(kind_phys), parameter :: rn_dst2 = 0.717e-6_kind_phys
+  real(kind_phys), parameter :: rn_dst3 = 1.576e-6_kind_phys
+  real(kind_phys), parameter :: rn_dst4 = 3.026e-6_kind_phys
+
+contains
+
+!> \section arg_table_dust_default_radii_timestep_init Argument Table
+!! \htmlinclude dust_default_radii_timestep_init.html
+  subroutine dust_default_radii_timestep_init(ncol, pver, ndust, rndst, errmsg, errflg)
```
nusbaume: This subroutine can be labeled `pure`.

schemes/microp_aero/scale_subgrid_vertical_velocity.F90
```diff
@@ -0,0 +1,72 @@
+! Apply minimum, maximum, and scaling to the subgrid vertical velocity.
+! wsub is the subgrid vertical velocity for droplet (liquid) nucleation.
+! wsubi is the subgrid vertical velocity for ice nucleation,
+! derived here from the raw wsub before wsub is scaled
+! (and capped at 0.2 m s-1 when preexisting ice is not used).
+module scale_subgrid_vertical_velocity
+
+  use ccpp_kinds, only: kind_phys
```
nusbaume: Move this `use` statement down to the subroutine level?

schemes/microp_aero/scale_subgrid_vertical_velocity.F90
```diff
@@ -0,0 +1,72 @@
...
+  implicit none
+  private
+
+  public :: scale_subgrid_vertical_velocity_run
+
+contains
+
+!> \section arg_table_scale_subgrid_vertical_velocity_run Argument Table
+!! \htmlinclude scale_subgrid_vertical_velocity_run.html
+  subroutine scale_subgrid_vertical_velocity_run( &
```
nusbaume: This subroutine can be labeled `pure`.

schemes/microp_aero/scale_subgrid_vertical_velocity_namelist.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0"?>
+
+<?xml-stylesheet type="text/xsl"?>
```
nusbaume, line 2: Add standard namelist comment block here?

schemes/microp_aero/scale_subgrid_vertical_velocity_namelist.xml
```diff
@@ -0,0 +1,73 @@
...
+  </entry>
+
+  <entry id="microp_aero_wsubi_scale">
+    <type>real</type>
+    <category>microphys</category>
+    <group>wsub_scale_nl</group>
+    <standard_name>tunable_parameter_for_subgrid_vertical_velocity_scaling_for_ice_nucleation</standard_name>
+    <units>1</units>
+    <desc>
+      Unitless scaling factor for the ice droplet subgrid scale vertical velocity
```
nusbaume: Nit-pick, but maybe better to say "crystal" instead of "droplet" here?


```suggestion
      Unitless scaling factor for the ice crystal subgrid scale vertical velocity
```

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+!  freezing on mineral dust (soot disabled) in cirrus clouds, and
+!  Meyers et al. (1992) deposition nucleation in mixed-phase clouds
+!
+!  The effect of preexisting ice crystals on ice nucleation in cirrus clouds is included,
+!  and also consider the sub-grid variability of temperature in cirrus clouds,
+!  following X. Shi et al. ACP (2014).
+!
+!  Ice nucleation in mixed-phase clouds now uses classical nucleation theory (CNT),
+!  follows Y. Wang et al. ACP (2014), Hoose et al. (2010).
+!
```
nusbaume, line 24: Can we add DOIs for all of the papers listed here?  I _believe_ they are the following:

```
Liu & Penner, 2005: 10.1127/0941-2948/2005/0059
Liu et al., 2007:   10.1175/JCLI4264.1
Shi et al., 2014:   10.5194/acp-15-1503-2015
Wang et al., 2014:  10.5194/acp-14-10411-2014
Hoose et al., 2010: 10.1175/2010JAS3425.1
```

For whatever reason the Meyers et al., 1992 doesn't have a DOI (at least that I could find), so maybe just include it's full citation here instead:

Meyers, M. P., P. J. DeMott, and W. R. Cotton (1992), New primary ice nucleation parameterizations in an explicit cloud model, J. Appl. Meteorol., 31, 708–721.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+!  With help from C. C. Chen and B. Eaton (2014)
+!-------------------------------------------------------------------------------
+
+use wv_saturation,  only: svp_water, svp_ice
+use ccpp_kinds,     only: kind_phys
+
+implicit none
+private
+save
```
nusbaume: Remove un-needed `save` here?

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+!  follows Y. Wang et al. ACP (2014), Hoose et al. (2010).
+!
+! Authors:
+!  Xiaohong Liu, 01/2005, modifications by A. Gettelman 2009-2010
+!  Xiangjun Shi & Xiaohong Liu, 01/2014.
+!
+!  With help from C. C. Chen and B. Eaton (2014)
+!-------------------------------------------------------------------------------
+
+use wv_saturation,  only: svp_water, svp_ice
```
nusbaume: I might move this `use` statement down to the relevant subroutine level.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+   real(kind_phys), intent(out) :: wpice      ! diagnosed Vertical velocity Reduction caused by preexisting ice (m/s), at Shom
+   real(kind_phys), intent(out) :: weff       ! effective Vertical velocity for ice nucleation (m/s); weff=wbar-wpice
+   real(kind_phys), intent(out) :: fhom       ! how much fraction of cloud can reach Shom
+   real(kind_phys), intent(out) :: regm       ! nucleation regime indiator
+   real(kind_phys), intent(out) :: oso4_num   ! so4 aerosol number (#/cm^3)
+   real(kind_phys), intent(out) :: odst_num   ! total dust aerosol number (#/cm^3)
+   real(kind_phys), intent(out) :: osoot_num  ! soot (hydrophilic) aerosol number (#/cm^3)
+
+   ! Optional Arguments
+   logical,  intent(in), optional :: call_frm_zm_in ! true if called from ZM convection scheme
```
nusbaume, line 136: You should certainly double-check me, but is `call_frm_zm_in` ever not `.false.` in CAM?  If so then I might vote to remove it (or add a note in the "future work" issue to remove it later).

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+     add_preexisting_ice = .true.
+   end if
+
+   if (use_preexisting_ice .and. (.not. call_frm_zm)) then
+
+      Ni_preice = ni_in*rhoair                    ! (convert from #/kg -> #/m3)
+      Ni_preice = Ni_preice / max(mincld,cldn)   ! in-cloud ice number density
+
+      if (Ni_preice > 10.0_kind_phys .and. qi > 1.e-10_kind_phys) then    ! > 0.01/L = 10/m3
+         Shom = -1.5_kind_phys   ! if Shom<1 , Shom will be recalculated in SUBROUTINE Vpreice, according to Ren & McKenzie, 2005
```
nusbaume: Include DOI for Ren & McKenzie paper?


```suggestion
         Shom = -1.5_kind_phys   ! if Shom<1 , Shom will be recalculated in SUBROUTINE Vpreice, according to Ren & McKenzie, 2005, DOI: 10.1256/qj.04.126
```

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+   esi   = 0._kind_phys
+   regm  = 0._kind_phys
+
+   oso4_num  = 0._kind_phys
+   odst_num  = 0._kind_phys
+   osoot_num = 0._kind_phys
+
+   if ((so4_num >= 1.0e-10_kind_phys .or. (soot_num+dst_num) >= 1.0e-10_kind_phys) .and. cldn > 0._kind_phys) then
+
+      if (RHimean.ge.1.2_kind_phys) then
```
nusbaume: Replace the older Fortran logical syntax with modern syntax here and below?  For example:


```suggestion
      if (RHimean >= 1.2_kind_phys) then
```

This can be added to the "future work" issue if we want to hold off for now, as it is just a formatting preference.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+               B = -10.41_kind_phys  * log(soot_num+dst_num) - 67.69_kind_phys
+               regm = A * log(wbar1) + B
+            end if
+
+            ! heterogeneous nucleation only
+            if (tc .gt. regm .or. so4_num < 1.0e-10_kind_phys) then
+
+               if(tc.lt.-40._kind_phys .and. wbar1.gt.1._kind_phys .and. so4_num >= 1.0e-10_kind_phys) then ! exclude T<-40 & W>1m/s from hetero. nucleation
+
+                  call hf(tc,wbar1,relhum*subgrid,so4_num,nihf)
```
nusbaume: Should we change `hf` to a more descriptive name, for example `homogeneous_freezing`?  This can wait for future work if preferred.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+                  niimm=0._kind_phys
+                  nidep=0._kind_phys
+
+                  ! If some homogeneous nucleation happened, assume all of the that heterogeneous
```
nusbaume: Grammar fix:


```suggestion
                  ! If some homogeneous nucleation happened, assume all of the heterogeneous
```

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+                     oso4_num  = nihf
+                  endif
+
+                  nihf      = nihf * fhom
+                  oso4_num  = oso4_num * fhom
+
+                  n1        = nihf + niimm
+               else
+
+                  call hetero(tc,wbar2,soot_num+dst_num,niimm,nidep)
```
nusbaume: Make this subroutine name more descriptive as well, for example `heterogeneous_freezing`?

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+
+            ! homogeneous nucleation only
+            else if (tc.lt.regm-5._kind_phys .or. (soot_num+dst_num) < 1.0e-10_kind_phys) then
+
+               call hf(tc,wbar1,relhum*subgrid,so4_num,nihf)
+               niimm=0._kind_phys
+               nidep=0._kind_phys
+
+               ! If some homogeneous nucleation happened, assume all of the that
+               ! heterogeneous and coarse mode sulfate particles nucleated.
```
nusbaume: grammar fix here as well:


```suggestion
               ! If some homogeneous nucleation happened, assume all of the
               ! heterogeneous and coarse mode sulfate particles nucleated.
```

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+            else
+
+               if (tc.lt.-40._kind_phys .and. wbar1.gt.1._kind_phys) then ! exclude T<-40 & W>1m/s from hetero. nucleation
+
+                  call hf(tc, wbar1, relhum*subgrid, so4_num, nihf)
+                  niimm = 0._kind_phys
+                  nidep = 0._kind_phys
+
+                  ! If some homogeneous nucleation happened, assume all of the
+                  ! that heterogeneous and coarse mode sulfate particles nucleated.
```
nusbaume: Another grammar fix (same as the previous ones):


```suggestion
                  ! If some homogeneous nucleation happened, assume all of the
                  ! heterogeneous and coarse mode sulfate particles nucleated.
```

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+                  if (nihf.gt.1e-3_kind_phys) then ! hom occur,  add preexisting ice
+                     niimm     = dst_num + soot_num       ! assuming dst_num freeze firstly
+                     odst_num  = dst_num
+                     osoot_num = soot_num
+
+                     oso4_num  = nihf
+                  endif
```
nusbaume: Total nit-pick, but can we convert all of the `endif` to `end if` in this subroutine?  Right now we seem to have a mix of both styles.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+   nuci   = nuci*1.e+6_kind_phys/rhoair    ! change unit from #/cm3 to #/kg
+   onimey = nimey*1.e+6_kind_phys/rhoair
+   onidep = nidep*1.e+6_kind_phys/rhoair
+   oniimm = niimm*1.e+6_kind_phys/rhoair
+   onihf  = nihf*1.e+6_kind_phys/rhoair
+end subroutine nucleati
+
+!===============================================================================
+
+subroutine hetero(T,ww,Ns,Nis,Nid)
```
nusbaume: This subroutine could be labeled `pure`.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+      Nis = exp(A22) * Ns**B22 * exp(B*T) * ww**C
+      Nis = min(Nis,Ns)
+
+      Nid = 0.0_kind_phys    ! don't include deposition nucleation for cirrus clouds when T<-37C
+
+end subroutine hetero
+
+!===============================================================================
+
+subroutine hf(T,ww,RH,Na,Ni)
```
nusbaume: This subroutine could be labeled `pure`.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+
+        endif
+
+      end if
+
+end subroutine hf
+
+!===============================================================================
+
+SUBROUTINE Vpreice(P_in, T_in, R_in, C_in, S_in, V_out)
```
nusbaume: This subroutine can be labeled `pure` (or I guess `PURE`).  In a perfect world it would probably be better as a function given that it has only one output, but that can be a future work issue.

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+
+   !  based on  Karcher et al. (2006)
```
nusbaume: I would include the DOI for the Karcher et al., 2006 reference, which I _think_ is the following:


```suggestion
   !  based on  Karcher et al. (2006), https://doi.org/10.1029/2005JD006219
```

schemes/microp_aero/nucleate_ice.F90
```diff
@@ -0,0 +1,589 @@
...
+   A3    = FA3c * T_1 / P
+   B1    = FLUX * SVOLc * CISAT * ( S-1.0_kind_phys )
+   B2    = FLUX * FDc * P * T_1**1.94_kind_phys
+   DLOSS = FPIVOLc * C * B1 * R**2 / ( 1.0_kind_phys+ B2 * R )
+   VICE  = ( A2 + A3 * S ) * DLOSS / ( A1 * S )  ! 2006,(19)
+   V_out = VICE*1e-2_kind_phys  ! cm/s => m/s
+
+END SUBROUTINE Vpreice
+
+subroutine frachom(Tmean,RHimean,detaT,fhom)
```
nusbaume: This subroutine can be labeled `pure`.  It would also probably be better  as a function given that it has only one output, but that can be a future-work issue.

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+  integer :: idxdst3 = -1 ! index in aerosol list for dust3
+  integer :: idxdst4 = -1 ! index in aerosol list for dust4
+
+contains
+
+  ! Initialize constants for droplet activation by bulk aerosols
+  subroutine ndrop_bam_init(amIRoot, iulog, mwh2o, r_universal, tmelt, rhoh2o, &
+    naer_all_out)
+
+    use shr_spfn_mod, only: erf => shr_spfn_erf
```
nusbaume, line 63: I will likely bring this up at a CSEG meeting, but I think for portability we should eventually try replacing this code here and below with the intrinsic `erf` function, as I suspect at this point most (all?) compilers support it.  Of course this may very likely change answers, so we should only worry about it when we get to the stage of finishing the "final" CCPP-ization steps.  Maybe add it to the "future work" issue so that we don't forget it?

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
+! Droplet activation by bulk aerosols.
+module ndrop_bam
```
nusbaume, line 2: Also forgot to say that I think renaming this to something like `droplet_activation_bam` at some point might be a better name, as I definitely remember scientists being confused in the past as to what "ndrop" was.  Of course since this won't actually impact the results at all this name change could be added to a future issue instead of worrying about it now.

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+      naer_all = 0
+    end if
+
+    allocate ( &
+      aername(naer_all), &
+      dryrad_aer(naer_all), &
+      density_aer(naer_all), &
+      hygro_aer(naer_all), &
+      dispersion_aer(naer_all), &
+      num_to_mass_aer(naer_all))
```
nusbaume: Check `allocate` status here?

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+    class(aerosol_properties), pointer :: aero_props_bam
+    !-------------------------------------------------------------------------------
+
+    ! Access the physical properties of the bulk aerosols that are affecting the climate
+    ! by using the abstract aerosol properties interface.
+
+    ! Find BAM properties object from factory
+    aero_props_bam => null()
+    do iaermod = 1, aerosol_instances_get_num_models()
+      aero_props_bam => aerosol_instances_get_props(iaermod, 0)
```
nusbaume: What is the `0` argument here?  This is just for my personal understanding, thanks!

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+    allocate ( &
+      aername(naer_all), &
+      dryrad_aer(naer_all), &
+      density_aer(naer_all), &
+      hygro_aer(naer_all), &
+      dispersion_aer(naer_all), &
+      num_to_mass_aer(naer_all))
+
+    do iaer = 1, naer_all
+      call aero_props_bam%get(iaer, 1, &
```
nusbaume: What is the `1` argument here?  This is also just for my personal understanding.  Thanks!

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+    pi = 4._kind_phys*atan(1.0_kind_phys) ! note: should use physconst pi in future cleanup.
+    sqpi = sqrt(pi)
+    surften = 0.076_kind_phys
+    aten = 2._kind_phys*mwh2o*surften/(r_universal*tmelt*rhoh2o)
+    alogaten = log(aten)
+    alog2 = log(2._kind_phys)
+    alog3 = log(3._kind_phys)
+    super(:) = 0.01_kind_phys*supersat(:)
+
+    allocate ( &
```
nusbaume, line 165: Check `allocate` status?

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+    real(kind_phys), intent(in)  :: qsmall_in          ! minimum cloud liquid threshold
+    real(kind_phys), intent(in)  :: ast(:, :)          ! stratiform_cloud_area_fraction [fraction]
+    real(kind_phys), intent(in)  :: numliq(:, :)       ! droplet number (#/kg)
+    real(kind_phys), intent(in)  :: deltatin           ! timestep (s)
+    real(kind_phys), intent(out) :: npccn(:, :)        ! droplet number tendency (#/kg/s)
+    real(kind_phys), intent(out) :: nacon(:, :, :)      ! contact freezing dust (#/m3), (ncol,pver,4)
+    real(kind_phys), intent(out) :: ccn(:, :, :)        ! CCN at 6 supersaturations (#/cm3), (ncol,pver,psat)
+    real(kind_phys), intent(out) :: naer2_diag(:, :, :) ! aerosol number conc (#/m3), (ncol,pver,naer_all)
+
+    character(len=512), intent(out) :: errmsg
```
nusbaume: I believe this should be `len=*` (and if that doesn't work in CAM now we'll need to change it when we eventually CCPP-ize this scheme):




```suggestion
    character(len=*), intent(out) :: errmsg
```

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+    real(kind_phys), intent(in) :: rair          ! dry air gas constant (J/K/kg)
+    real(kind_phys), intent(in) :: tmelt         ! freezing point of water (K)
+    real(kind_phys), intent(in) :: cpair         ! specific heat of dry air (J/K/kg)
+    real(kind_phys), intent(in) :: rh2o          ! water vapor gas constant (J/K/kg)
+    real(kind_phys), intent(in) :: rhoh2o        ! density of liquid water (kg/m3)
+    real(kind_phys), intent(in) :: latvap        ! latent heat of vaporization (J/kg)
+
+    ! output
+    real(kind_phys), intent(out) :: nact         ! number fraction of aerosols activated
+    character(len=512), intent(out) :: errmsg
```
nusbaume: Make this `len=*`?


```suggestion
    character(len=*), intent(out) :: errmsg
```

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+    if (wbar .le. 0._kind_phys) return
+
+    pres = rair*rhoair*tair
+    diff0 = 0.211e-4_kind_phys*(pref/pres)*(tair/tmelt)**1.94_kind_phys
+    conduct0 = (5.69_kind_phys + 0.017_kind_phys*(tair - tmelt))*4.186e2_kind_phys*1.e-5_kind_phys ! convert to J/m/s/deg
+    call qsat(tair, pres, es, qs)
+    dqsdt = latvap/(rh2o*tair*tair)*qs
+    alpha = gravit*(latvap/(cpair*rh2o*tair*tair) - 1._kind_phys/(rair*tair))
+    gammaloc = (1 + latvap/cpair*dqsdt)/(rhoair*qs)
+    ! growth coefficent Abdul-Razzak & Ghan 1998 eqn 16
```
nusbaume: I might add the DOI here:


```suggestion
    ! growth coefficent Abdul-Razzak & Ghan 1998 eqn 16, https://doi.org/10.1029/97JD03735
```

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+      return
+    end if
+
+    nact = 0._kind_phys
+
+    if (nmode .eq. 1 .and. na(1) .lt. 1.e-20_kind_phys) return
+
+    if (wbar .le. 0._kind_phys) return
+
+    pres = rair*rhoair*tair
```
nusbaume, line 445: I wouldn't do anything about this now (as it would change answers), but it makes me a little nervous that this is calculating its own pressure instead of just using the provided pressure (e.g. `air_pressure`).  

At some point it might be good to double-check the paper and these equations to make sure everything is consistent with the rest of the physics (e.g. it's not using dry air pressure when expecting total air pressure), but for now maybe just add it to the "future work" issue?

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+      etafactor2, &
+      amcubeloc, &
+      lnsmloc)
+
+  end subroutine activate
+
+!===============================================================================
+
+  subroutine ccn_diag(ncol, pver, top_lev, maerosol, naer2, ccn_out)
+    use shr_spfn_mod, only: erfc => shr_spfn_erfc
```
nusbaume, line 523: At some point we'll likely want to replace this with the intrinsic `erfc` function (but for now we can just add it to the future work issue).

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+
+        end do   ! supersaturation
+      end do      ! bulk aerosol
+    end do         ! level
+
+  end subroutine ccn_diag
+
+!===============================================================================
+
+  subroutine maxsat(zeta, eta, nmode, smc, smax)
```
nusbaume: This subroutine can be labeled `pure` (and ideally should be a function, as it has only a single output, but that can be a future-work issue).

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -0,0 +1,636 @@
...
+    integer :: m  ! mode index
+    real(kind_phys) :: sum, g1, g2
+
+    do m = 1, nmode
+      if (zeta .gt. 1.e5_kind_phys*eta(m) .or. smc(m)*smc(m) .gt. 1.e5_kind_phys*eta(m)) then
+        ! weak forcing. essentially none activated
+        smax = 1.e-20_kind_phys
+      else
+        ! significant activation of this mode. calc activation all modes.
+        go to 1
```
nusbaume: You can probably guess what I am going to ask for here.

nusbaume (changes requested):
Thanks for this initial effort in bringing over the aerosol microphysics @jimmielin!  I realize I have a bunch of questions and change requests, but the vast majority can be saved for a future PR (I just wanted to make sure I noted them here).

Of course if you have any questions or concerns with any of my requests just let me know.  Thanks again!

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
jimmielin: Thanks, marked as pure

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
jimmielin: Added comment
```
    ! Assuming isotropic turbulent motion, the vertical velocity variance is
    ! two thirds of the TKE. Cap the sub-grid vertical velocity at 10 m s-1.
```

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
jimmielin: Thanks, added!

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
jimmielin: Yes, it is what CAM is doing now and it is correct. The port is bit-for-bit in CAM, using the latest CLUBB.

clubb_intr has a bracket `if(clubb_l_ascending_grid)` enclosing `wp2_pbuf(:,nzm_clubb:1:-1)` flips and a flip back afterwards so the pbuf field is always top-down: if CLUBB's upside-down flip propagated to `pbuf` consumers then the CLUBB PR would have to send that flag to all downstream users to flip their pbuf fields when retrieving which wouldn't make sense.

But you got me on a deeper bug: `wp2` is a partial array up to `nzm_clubb` but I declared it as `vertical_interface_dimension` in the registry (it is written like that in the snapshot, but that is only because CAM7 LT is not tall enough so CLUBB goes up to 1.)

I will make the proper fix to dimensions, it needs a patch in CAM and a cautionary issue in atmospheric_physics that when CLUBB is ported `wp2` must be a full `ilev` field and not dimensioned `nzm_clubb`. 

I cannot make this array `nzm_clubb` here since CLUBB is not CCPPized, I have no way to determine the dimension.

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
jimmielin: Moved!

schemes/microp_aero/dust_default_radii.F90
jimmielin: Purified!

schemes/microp_aero/scale_subgrid_vertical_velocity.F90
jimmielin: Moved!

schemes/microp_aero/scale_subgrid_vertical_velocity.F90
jimmielin: Purified!

schemes/microp_aero/scale_subgrid_vertical_velocity_namelist.xml
jimmielin: Added!

schemes/microp_aero/scale_subgrid_vertical_velocity_namelist.xml
jimmielin: Crystallized!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks for chasing down the DOIs. Meyers is [https://doi.org/10.1175/1520-0450(1992)031<0708:NPINPI>2.0.CO;2](https://doi.org/10.1175/1520-0450(1992)031%3C0708:NPINPI%3E2.0.CO;2), so I added that too.

schemes/microp_aero/nucleate_ice.F90
jimmielin: Removed! I will double check with the CAM regression tests if this is happy in CAM

schemes/microp_aero/nucleate_ice.F90
jimmielin: Moved to `nucleati`!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, added!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, fixed!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, fixed!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, fixed!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Purified!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Purified!

schemes/microp_aero/nucleate_ice.F90
jimmielin: PURIFIED! Actually, I made it not scream at us, so purified. I kept the body as-is.

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, added!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Purified! I will add this to a future work issue

schemes/microp_aero/ndrop_bam.F90
jimmielin: I think this will happen a lot throughout all the aerosol code, so I will make a generic "replace shr_spfn_erf with erf intrinsic" issue instead of adding to the issue for this one. There's at least a dozen of these in MAM.

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks for the suggestion. Eventually the plan is to get rid of this module in favor of unifying all paths for aerosol activation, especially with `ndrop` (the MAM/CARMA path). 

My current plan for MAM is `aero_activate` as the portable core extracted from `ndrop`, it contains `activate_aerosol` (called by the convective/wetdep) but also confusingly `dropmixnuc` (that one doesn't roll off the tongue as well as `ndrop`) which is a scheme that needs to be CCPPized too. I would say we discuss the naming when all the pieces are ready for renaming (fun fact: did you know that `rename` is also the name of a CCPP scheme for modal aerosols? :-) )

schemes/microp_aero/ndrop_bam.F90
jimmielin: Done and threaded through the CCPP init

schemes/microp_aero/ndrop_bam.F90
jimmielin: It's the index for the climate list of aerosols for everything aerosol (to be distinguished with the >0 values which are for diagnostic lists). It is indexed in `radiative_aerosol_definitions`. I've made this a keyword argument, I think it's less cryptic than just `, 0)`.

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks for the question. In bulk aerosols every bulk aerosol is a bin and each bin has exactly one species (the bulk aerosol itself). I've made this a keyword argument to make it less cryptic. It is especially so for BAM. `bulk_aerosol_properties_mod.F90` in BAM (CAM-SIMA PR or pre-existing in CAM) has a brief comment on this, since I've probably asked Francis this exact question more times than I'd like to admit.

schemes/microp_aero/ndrop_bam.F90
jimmielin: Checked and threaded through

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks, `*` throughout

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks, `*` throughout

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks, added!

schemes/microp_aero/ndrop_bam.F90
jimmielin: Purified!

schemes/microp_aero/ndrop_bam.F90
jimmielin: Flattened, thanks!

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks, added https://github.com/ESCOMP/atmospheric_physics/issues/418

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, replaced throughout

schemes/microp_aero/nucleate_ice.F90
jimmielin: Good suggestion, this makes it possible to be consistent with `hf` too, changed!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, changed!

schemes/microp_aero/nucleate_ice.F90
jimmielin: Thanks, fixed throughout.

schemes/microp_aero/nucleate_ice.F90
jimmielin: Added to future issue!

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks, added https://github.com/ESCOMP/atmospheric_physics/issues/421

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
nusbaume: Thanks for diving into this @jimmielin!  I am happy with the changes in this PR, but long-term we'll definitely want `wp2` to be dimensioned `nzm_clubb`, as otherwise we'll be wasting a lot of memory when running with high-top (e.g. WACCM) configurations.  

So, how should we best keep track of this to make sure this dimension change is brought into CAM and SIMA when CLUBB is ready?  I am happy to make an issue, but I'm also aware that issues can get lost in the noise.  Alternatively we can convert these CCPP-schemes into simple helper functions, in which case we don't need to specify the metadata at all, but that would take more work.  We could also just create a new `nzm_clubb` dimension variable in the SIMA registry that temporarily matches the correct size, and that will be marked as replaceable once CLUBB come in.  Any thoughts?

Also pinging @bstephens82 in case he has any preferences, given that he is helping to bring in CLUBB.

jimmielin: Thanks @nusbaume. I had opened an issue https://github.com/ESCOMP/atmospheric_physics/issues/419. It can be changed in the other direction or go in a different direction when we get there.

It cannot be in the registry because it depends on `trop_cloud_top_lev` and `pver(p)`. The template to bring in the `nzm_clubb` dimension is already in UW PBL with the `ncvmax` dimension, which is to make it a runtime output from the initialization. 

In my opinion it's not helpful to speculatively build the dimension for CLUBB before CLUBB is brought in, it will only create headaches for merging and it might not be the right shape that the scheme needs, which will only be apparent when the scheme itself is brought in. When CLUBB is brought in all the downstream consumers of CLUBB inputs have to be updated anyway, it is more efficient to work like this (e.g., Beljaars replacing the stub). We can only reason so much in the abstract.

schemes/microp_aero/ndrop_bam.F90
nusbaume: Thanks for the explanation @jimmielin!  I'm happy to leave this as-is for now, and then tackle it when you are ready to implement MAM.  I'll go ahead and mark this particular concern as resolved.

schemes/microp_aero/compute_subgrid_vertical_velocity.F90
nusbaume: Whoops, didn't see the issue!   I'll move the conversation over there (I'll also leave this open just for posterity, but will consider it resolved for this PR).

schemes/microp_aero/ndrop_bam.F90
```diff
@@ -598,20 +610,22 @@ subroutine maxsat(zeta, eta, nmode, smc, smax)
...
     real(kind_phys) :: sum, g1, g2
+    logical :: weak_forcing
 
+    weak_forcing = .true.
     do m = 1, nmode
-      if (zeta .gt. 1.e5_kind_phys*eta(m) .or. smc(m)*smc(m) .gt. 1.e5_kind_phys*eta(m)) then
-        ! weak forcing. essentially none activated
-        smax = 1.e-20_kind_phys
-      else
+      if (.not. (zeta .gt. 1.e5_kind_phys*eta(m) .or. smc(m)*smc(m) .gt. 1.e5_kind_phys*eta(m))) then
```
nusbaume: Nit-pick, but change `.gt.` to `>`?

nusbaume (approved):
Thanks for addressing my concerns @jimmielin!  I have one last nit-pick request, and also left a few comments unresolved in case we want to add them to issues (although it's not required).  Otherwise I think this PR is ready to go!  Thanks again!

schemes/microp_aero/ndrop_bam.F90
jimmielin: Thanks, done!

jimmielin:
Pending a test run with CAM regression tests to make sure everything is still working.

PR #416, opened 2026-06-28
Add XML schema validation workflow

### LINKED ISSUE #59: Add xmllint Github Action
Given that there will soon be multiple XML files (of at least two different types) in this repo that host models will depend on, it could help to implement an automatic `xmllint` Github Action that ensures that any new XML file validates correctly. 

This could likely just be a copy of the code eventually used for the same action described in NCAR/CAMDEN#188, except with the registry schema replaced with the CCPP SDF schema.

virtuousluna:
Tag name: xml-validation
Originator(s): Rosio Rodriguez

_Description (include issue title and the keyword ['closes', 'fixes', 'resolves'] and issue number):_ Resolves #59.

This PR adds a GitHub Actions workflow for XML schema validation in atmospheric_physics. The workflow uses xmllint to validate hand-written XML files in the repository.

_The workflow validates two categories of XML files:_

Suite definition files in suites/ and test/test_suites/, using the CCPP suite schema from the ccpp-framework repository.
Scheme namelist XML files under schemes/, using the CIME ParamGen entry_id_pg schema.

This adds automated CI coverage so XML schema issues can be caught during pull requests instead of being discovered later during model configuration or build workflows.

_List all namelist files that were added or changed:_

None. This PR does not add or modify any namelist XML files. It only adds validation coverage for existing scheme namelist XML files.

_List all files eliminated and why:_ None.

_List all files added and what they do:_ .github/workflows/xml_validation.yaml

Adds a GitHub Actions workflow named XML schema validation. The workflow installs xmllint, checks out the needed schema sources from ccpp-framework and CIME, validates suite definition XML files, and validates all scheme namelist XML files matching *_namelist.xml under the schemes/ directory.

_List all existing files that have been modified, and describe the changes:_
(Helpful git command: git diff --name-status main...<your_branch_name>)

None. This PR adds a new workflow file and does not modify existing source, namelist, or physics files.

_List all automated tests that failed, as well as an explanation for why they weren't fixed:_ None in the final PR state.

During development, I caused the XML schema validation workflow to fail by temporarily changing the root tag in one scheme namelist XML file. This was done to confirm that the workflow correctly catches invalid XML. The file was restored, and the workflow passed successfully afterward.

_Is this an answer-changing PR? If so, is it a new physics package, algorithm change, tuning change, etc?_

No. This is not an answer-changing PR. It only adds CI validation infrastructure for XML files.

_If yes to the above question, describe how this code was validated with the new/modified features:_

Not applicable because this is not an answer-changing PR.

_The workflow was validated by running it in a fork and confirming the expected behavior:_

- valid XML files passed schema validation
- a deliberately broken scheme namelist XML file failed schema validation
- restoring the XML file caused the workflow to pass again


.github/workflows/xml_validation.yaml
```diff
@@ -0,0 +1,63 @@
...
+on:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  validate-xml:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout atmospheric_physics
+        uses: actions/checkout@v4
```
nusbaume: I believe this can now be bumped up to version 7:


```suggestion
        uses: actions/checkout@v7
```

.github/workflows/xml_validation.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Install xmllint
+        run: sudo apt-get update && sudo apt-get install -y libxml2-utils
+
+      # The suite schema lives in the ccpp-framework repository.
+      # Pin the same tag that CAM-SIMA pins (see CAM-SIMA/.gitmodules)
```
nusbaume: Instead of pinning to the same tags that CAM-SIMA uses, I would just check out the head of both the ccpp-framework and CIME repos.  This would not only help reduce the maintenance burden of needing to keep this workflow file in-sync with CAM-SIMA, but would also then act as a "canary" test, and allow us to catch any upstream changes to the schema files before they cause a real problem (although to be honest the schema files don't really change that often).

nusbaume (changes requested):
This is excellent work @virtuousluna!  I just had a couple of requests, but hopefully nothing too problematic (and if so please let me know).  

Also, if you could make sure that the action fails when a suite definition XML file is incorrectly coded as well that would be helpful as a way to make sure we have both possible file types tested properly.  Thanks!

.github/workflows/xml_validation.yaml
virtuousluna: Thanks for the review. I updated the workflow to use `actions/checkout@v7`, removed the pinned `ccpp-framework` and CIME tags so the workflow checks out the current repository heads, and updated the comments accordingly.

I also tested the suite definition XML validation path by temporarily breaking `suites/suite_adiabatic.xml`. The XML schema validation workflow failed as expected. I then restored the file, and the XML schema validation workflow passed again.

nusbaume (approved):
Thanks for implementing my requests @virtuousluna!  Everything looks great to me now.

jimmielin (approved):
Thanks @virtuousluna! I am happy to bring this in shortly

PR #422, opened 2026-07-13
Add RRTMGP and surface coupling to convection-permitting suite

kuanchihwang:
### Tag name

None (Bit-for-bit)

### Originator(s)

kuanchihwang

### Description (include the issue title, and the keyword ['closes', 'fixes', 'resolves'] followed by the issue number)

This PR adds RRTMGP and surface coupling to the convection-permitting suite. It also applies a blanket fix to the inconsistent tendency handling across various MMM physics schemes.

### List all namelist files that have been modified

None

### List all files eliminated and why

None

### List all files added and what they do

None

### List all existing files that have been modified, and describe the changes

```
M       schemes/mmm/bl_gwdo_compat.meta
    * Fix inconsistent tendency behaviors in MMM physics schemes
M       schemes/mmm/bl_mynn_compat.meta
M       schemes/mmm/cu_ntiedtke_compat.meta
M       schemes/mmm/mmm_physics_compat.F90
    * Update standard names and add interstitial for RRTMGP
    * Fix inconsistent tendency behaviors in MMM physics schemes
M       schemes/mmm/mmm_physics_compat.meta
    * Update standard names and add interstitial for RRTMGP
    * Fix metadata
    * Fix inconsistent tendency behaviors in MMM physics schemes
M       test/test_suites/suite_convection_permitting.xml
    * Add RRTMGP to convection-permitting suite
    * Update standard names and add interstitial for RRTMGP
    * Enable surface coupling
    * Fix inconsistent tendency behaviors in MMM physics schemes
M       test/unit-test/tests/mmm/mmm_physics_compat_tests.pf
    * Add unit tests
```

### List all automated tests that failed, as well as an explanation for why they were not fixed

All passed.

### Is this an answer-changing PR? If so, is it a new physics package, algorithm change, tuning change, etc?

Yes, but only for the convection-permitting suite.

### If yes to the above question, describe how this code was validated with the new/modified features

Configure, build, and run CAM-SIMA with the `MPASCP` compset.


test/test_suites/suite_convection_permitting.xml
```diff
@@ -20,6 +22,56 @@
 
     <scheme>mmm_physics_compat</scheme>
 
+    <!-- Begin section RRTMGP -->
```
nusbaume: I might note here that RRTMGP is the radiation scheme (as it may not be obvious to new users):


```suggestion
    <!-- Begin section RRTMGP (radiation scheme) -->
```

schemes/mmm/mmm_physics_compat.F90
```diff
@@ -249,6 +253,223 @@ pure subroutine mmm_physics_accumulate_tendencies_run( &
...
+
+        errmsg = ''
+        errflg = 0
+
+        ! Save pending tendencies to final tendencies.
+        !
+        ! The reason for this mechanism is to mitigate the inconsistent tendency behaviors across various MMM physics schemes.
+        ! Although some schemes correctly accumulate their tendencies with others, there are certain schemes that outright
+        ! overwrite the tendencies from others. Instead of wrestling with the problematic schemes individually like MPAS does,
+        ! which is very error-prone, a suite-level scheme is introduced to blanketly enforce the correct behavior.
```
nusbaume, line 413: I'm totally happy with this for now, but long-term is there a chance that we can eventually fix the problematic schemes so that this helper scheme is no longer needed?  Definitely not a high priority, but it does seem like a way to reduce the total amount of variables (and thus memory) that is needed for a simulation.

nusbaume (approved):
Thanks @kuanchihwang!  I just had one optional request and a comment covering the discussion we had in our meeting.  Otherwise these changes look good to me!

test/test_suites/suite_convection_permitting.xml
kuanchihwang: I updated the comments in https://github.com/ESCOMP/atmospheric_physics/pull/422/commits/1c1c32c74ed0def45a00172ef654d8406bbca2df.

schemes/mmm/mmm_physics_compat.F90
kuanchihwang: I can definitely move the fixes to the compatibility wrapper of each scheme if that is the desired direction, but we likely cannot modify the main schemes at MMM-physics because it will break the dependent MMM models.

MPAS also suffers from this issue: The `_p`-suffix variables can be found all over the place in each physics driver (e.g., the PBL physics driver<sup>[[1]](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.4.1/src/core_atmosphere/physics/mpas_atmphys_driver_pbl.F#L450-L527)</sup>).

schemes/mmm/mmm_physics_compat.F90
```diff
@@ -249,6 +253,223 @@ pure subroutine mmm_physics_accumulate_tendencies_run( &
...
+
+        rqsblten(:, :) = rqsblten(:, :) + rqsblten_p(:, :)
+
+        rozblten(:, :) = rozblten(:, :) + rozblten_p(:, :)
+
+        rnwfablten(:, :) = rnwfablten(:, :) + rnwfablten_p(:, :)
+        rnifablten(:, :) = rnifablten(:, :) + rnifablten_p(:, :)
+        rnbcablten(:, :) = rnbcablten(:, :) + rnbcablten_p(:, :)
+
+        ! After saving, zero out pending tendencies to make this subroutine idempotent.
```
peverwhee: This might be a bit redundant, but can we elaborate more on why this is important? Suggestion below.
```suggestion
        ! After saving, zero out pending tendencies to make this subroutine idempotent, guarding against repeated application of the same tendencies.
```

peverwhee (approved):
one tiny comment quibble but nothing that holds back my approval!

schemes/mmm/mmm_physics_compat.F90
kuanchihwang: I updated the code comments in https://github.com/ESCOMP/atmospheric_physics/pull/422/commits/869aa5839ff4936629d1bb79a4d59f6d355d0cff.

PR #424, opened 2026-07-18
Change prescribed chemistry/aerosol schemes to use ccpp_scheme_utils; fully use diagnostic names in sima_state_diagnostics

jimmielin:
Originator(s): @jimmielin 

- Change prescribed chemistry/aerosol schemes to use ccpp_scheme_utils so they are case-insensitive - companion PR to https://github.com/ESCOMP/CAM-SIMA/pull/518
- #397 (in 979c6bca02a4e4c76d94205c27831a7be66fb035) changed `sima_state_diagnostics` to use `%diagnostic_name` to add field but not the write loop, so if the two cases mismatched (due to the case-insensitivity fixes they now do) the `history_out_field` would be mismatched and would write uninitialized heap memory into the netCDF file rather than the variable.

```
M       schemes/chemistry/prescribed_aerosols.F90
M       schemes/chemistry/prescribed_aerosols.meta
M       schemes/chemistry/prescribed_ozone.F90
M       schemes/chemistry/prescribed_ozone.meta
  - change to use ccpp_scheme_utils.
 
M       schemes/sima_diagnostics/sima_state_diagnostics.F90
  - change to use diagnostic_name in history_out_field.
```

### DIFF
--- a/schemes/chemistry/prescribed_aerosols.F90
+++ b/schemes/chemistry/prescribed_aerosols.F90
@@ -515,7 +515,7 @@ contains
       ! used as a negative array subscript below.
       if (const_idx < 0) then
         errflg = 1
-        errmsg = subname // ': could not find constituent ' // trim(aero_map_list(i)%constituent_name)
+        errmsg = subname // ': could not find constituent "' // trim(aero_map_list(i)%constituent_name)//'"'
         return
       end if
 
--- a/schemes/chemistry/prescribed_aerosols.F90
+++ b/schemes/chemistry/prescribed_aerosols.F90
@@ -457,7 +457,6 @@ contains
 !! \htmlinclude prescribed_aerosols_run.html
   subroutine prescribed_aerosols_run( &
     ncol, pver, pcnst, &
-    const_props, &
     pi, &
     pmid, pint, phis, zi, & ! necessary fields for trcdata read.
     constituents, &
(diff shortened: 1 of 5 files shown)


climbfuji (commented):
looks like this will work fine with capgen v1, too

schemes/chemistry/prescribed_aerosols.F90
```diff
@@ -513,11 +507,18 @@ subroutine prescribed_aerosols_run( &
...
-           const_idx, errmsg, errflg)
+      call ccpp_constituent_index(trim(aero_map_list(i)%constituent_name), &
+           const_idx, errmsg=errmsg, errcode=errflg)
       if (errflg /= 0) return
 
+      ! Guard against an unresolved constituent index, which would otherwise be
+      ! used as a negative array subscript below.
+      if (const_idx < 0) then
+        errflg = 1
+        errmsg = subname // ': could not find constituent ' // trim(aero_map_list(i)%constituent_name)
```
nusbaume: I might quote-wrap the constituent name here, just in case the`constituent_name` is blank or has some other weird characters:


```suggestion
        errmsg = subname // ': could not find constituent "' // trim(aero_map_list(i)%constituent_name)//'"'
```

nusbaume (approved):
Thanks @jimmielin!  I just had one optional request.

cacraigucar (approved):
Looks good - interesting that ccpp_constituent_index does not return an errcode/errflag from one else section, so that you need to check for a negative const_idx after having checked for errocde.

PR #427, opened 2026-07-22
Export CO2 value to the land model via coupler

jimmielin:
Originator(s): @jimmielin assisted-by claude-fable:5

- Update the `set_surface_coupling_vars` scheme to pass the `prescribed_volume_mixing_ratio_of_co2` prescribed via `prescribe_radiative_gas_concentrations` to the coupler

CAM just does this in `camsrfexch.F90`
```fortran
cam_out%co2diag(:ncol) = chem_surfvals_get('CO2VMR') * 1.0e+6_r8
```

A       schemes/utilities/set_surface_coupling_vars_namelist.xml

```
M       schemes/utilities/set_surface_coupling_vars.F90
M       schemes/utilities/set_surface_coupling_vars.meta
  - update to set co2diag, co2prog
```


jimmielin:
Thanks @nusbaume for the comments - I have updated the PR to just use `set_surface_coupling_vars`!

schemes/utilities/set_surface_coupling_vars.F90
```diff
@@ -114,6 +130,31 @@ subroutine set_surface_coupling_vars_run(ncol, pver, ncnst, gravit, rair, phis,
      end do
    end do
 
+   ! Export the CO2 volume mixing ratio for surface models.
+   ! do_diagnostic_co2 selects whether it is sent in the diagnostic
+   ! or prognostic coupler field (for surface model).
+   co2diag(:) = -1._kind_phys
+   co2prog(:) = -1._kind_phys
```
nusbaume: I actually might initialize these variables to zero instead of negative one, as that feels like a more physically-appropriate setting (i.e. if there is no CO2 in the atmosphere then there is a concentration of 0 ppm).  I can always revert it if this causes downstream problems.


```suggestion
   co2diag(:) = 0._kind_phys
   co2prog(:) = 0._kind_phys
```

nusbaume (approved):
Thanks @jimmielin!  I have one change request but it doesn't require a re-review.

schemes/utilities/set_surface_coupling_vars.F90
jimmielin: I think -1 and 0 will crash equally downstream (it'll crash LUNA within CTSM after a few days) so I am okay with either value!

peverwhee (approved):

PR #428, opened 2026-07-27
atmos_phys0_29_000: Port conv_water_4rad to atmos_phys (convective_cloud_water) and cloud_water_paths

jimmielin:
Tag name (The PR title should also include the tag name): atmos_phys0_29_000
Originator(s): @jimmielin (co-authored with `claude-fable:5`)

- Ports CAM's `conv_water_4rad` (src/physics/cam/conv_water.F90) and 
- the two-moment cloud-water-path block of `cloud_diagnostics_calc` (src/physics/cam/cloud_diagnostics.F90), 
- plus the snow/graupel cloud-fraction and water-path block from micro_pumas_cam.

The first one is a portable core scheme reused in CAM.

A       schemes/cloud_water/convective_cloud_water_namelist.xml

```
A       schemes/cloud_water/cloud_water_paths.F90
A       schemes/cloud_water/cloud_water_paths.meta
  - new scheme

A       schemes/cloud_water/convective_cloud_water.F90
A       schemes/cloud_water/convective_cloud_water.meta
A       schemes/cloud_water/convective_cloud_water_namelist.xml
  - new scheme

A       schemes/sima_diagnostics/cloud_water_paths_diagnostics.F90
A       schemes/sima_diagnostics/cloud_water_paths_diagnostics.meta
A       schemes/sima_diagnostics/convective_cloud_water_diagnostics.F90
A       schemes/sima_diagnostics/convective_cloud_water_diagnostics.meta
  - diagnostics scheme.

A       test/test_suites/suite_cloud_water.xml
  - test SDF.

A       schemes/rasch_kristjansson/rk_stratiform_prepare_flux_for_check_energy.F90
A       schemes/rasch_kristjansson/rk_stratiform_prepare_flux_for_check_energy.meta
  - wire up RK for check energy independently of prec_sed 
```

```
M       suites/suite_cam5.xml
M       suites/suite_cam7.xml
   - put in production SDF

M       suites/suite_cam4.xml
   - add check_energy for RK
   NOTE: cam4 suite intentionally not updated for conv_water until the FHIST_C4 PR

M       schemes/rasch_kristjansson/prognostic_cloud_water.F90
M       schemes/rasch_kristjansson/prognostic_cloud_water.meta
M       schemes/rasch_kristjansson/rk_stratiform.meta
   - fix water budget leak due to rliq being subtracted from prec_sed
     that is passed to the coupler to the surface model

M       schemes/rrtmgp/rrtmgp_inputs.F90
   - fix missing _kind_phys
```

Yes
New physics package


jimmielin:
Uncomment PUMAS from CAM7 suite after conv_water_4rad is merged in

schemes/cloud_water/cloud_water_paths.F90
```diff
@@ -0,0 +1,171 @@
...
+    iclwp, iciwp, &
+    cldfsnow, icswp, &
+    cldfgrau, icgrauwp, &
+    icimr, icwmr, iwc, lwc, &
+    errmsg, errflg)
+
+    integer,         intent(in)  :: ncol
+    integer,         intent(in)  :: pver
+    integer,         intent(in)  :: top_lev           ! troposphere cloud physics top level [index]
+    integer,         intent(in)  :: conv_water_in_rad ! Use convective water in radiation? [flag]
```
nusbaume: I might update the comment here slightly to clarify why it's not a logical:


```suggestion
    integer,         intent(in)  :: conv_water_in_rad ! control variable, with zero meaning no convective water used [1]
```

schemes/cloud_water/cloud_water_paths.F90
```diff
@@ -0,0 +1,171 @@
...
+    icwmr = 0._kind_phys
+    iwc = 0._kind_phys
+    lwc = 0._kind_phys
+
+    do k = top_lev, pver
+      do i = 1, ncol
+        ! Limits for in-cloud mixing ratios consistent with MG microphysics
+        ! in-cloud mixing ratio maximum limit of 0.005 kg/kg
+        icimr(i, k) = min(allcld_ice(i, k)/max(0.0001_kind_phys, cld(i, k)), 0.005_kind_phys)
+        icwmr(i, k) = min(allcld_liq(i, k)/max(0.0001_kind_phys, cld(i, k)), 0.005_kind_phys)
```
nusbaume: Should we use the `mincld` parameter here?


```suggestion
        icimr(i, k) = min(allcld_ice(i, k)/max(mincld, cld(i, k)), 0.005_kind_phys)
        icwmr(i, k) = min(allcld_liq(i, k)/max(mincld, cld(i, k)), 0.005_kind_phys)
```

schemes/cloud_water/cloud_water_paths.F90
```diff
@@ -0,0 +1,171 @@
...
+            (ls_liq(i, k) < 1.e-10_kind_phys)) then
+          cldfgrau(i, k) = 0._kind_phys
+        end if
+        ! If no cloud and graupel, then set to 0.25
+        if ((cldfgrau(i, k) <= 1.e-4_kind_phys) .and. (qgout(i, k) > 1.e-9_kind_phys)) then
+          cldfgrau(i, k) = 0.25_kind_phys
+        end if
+
+        ! Calculate in-cloud graupel water path
+        icgrauwp(i, k) = qgout(i, k)/max(1.e-2_kind_phys, cldfgrau(i, k))*pdel(i, k)/gravit
```
nusbaume: I wonder if we should create a new parameter that makes it clear that graupel uses a different minimum cloud fraction relative to all of the other hydrometeors?

schemes/cloud_water/cloud_water_paths.meta
```diff
@@ -0,0 +1,183 @@
...
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = in
+[ iclwp ]
+  standard_name = in_cloud_liquid_water_path_for_radiation
+  units = kg m-2
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ iciwp ]
+  standard_name = in_cloud_ice_water_path
```
nusbaume, line 119: Just curious why `iclwp` has the `for_radiation` suffix, but `iciwp` doesn't?

schemes/cloud_water/cloud_water_paths.meta
```diff
@@ -0,0 +1,183 @@
...
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ cldfgrau ]
+  standard_name = liquid_plus_graupel_stratiform_cloud_area_fraction
+  units = fraction
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ icgrauwp ]
+  standard_name = stratiform_in_cloud_graupel_water_path
```
nusbaume, line 143: Why does this variable have the `stratiform` prefix, but none of the other `in_cloud_X_water_path` variables do?

schemes/cloud_water/convective_cloud_water.F90
```diff
@@ -0,0 +1,236 @@
...
+    integer,         intent(in)  :: conv_water_in_rad
+    real(kind_phys), intent(in)  :: frac_limit                  ! minimum cloud fraction [fraction]
+    ! one-moment (RK) microphysics: use rei unclamped in the ice absorptivity;
+    ! two-moment clamps rei to [13,130] micron
+    logical,         intent(in)  :: one_mom_clouds
+    real(kind_phys), intent(in)  :: gravit                      ! gravitational acceleration [m s-2]
+
+    real(kind_phys), intent(in)  :: pdel(:, :)                  ! (ncol,pver) moist pressure difference across layer [Pa]
+    real(kind_phys), intent(in)  :: ls_liq(:, :)                ! (ncol,pver) large-scale contributions to GBA cloud liq [kg kg-1]
+    real(kind_phys), intent(in)  :: ls_ice(:, :)                ! (ncol,pver) large-scale contributions to GBA cloud ice [kg kg-1]
```
nusbaume: I might include "stratiform" in the comments here as well:


```suggestion
    real(kind_phys), intent(in)  :: ls_liq(:, :)                ! (ncol,pver) large-scale (stratiform) contributions to GBA cloud liq [kg kg-1]
    real(kind_phys), intent(in)  :: ls_ice(:, :)                ! (ncol,pver) large-scale (stratiform) contributions to GBA cloud ice [kg kg-1]
```

schemes/cloud_water/convective_cloud_water.F90
```diff
@@ -0,0 +1,236 @@
...
+
+    real(kind_phys), intent(in)  :: sh_icwmr(:, :)              ! (ncol,pver) shallow conv. cloud water [kg kg-1]
+    real(kind_phys), intent(in)  :: dp_icwmr(:, :)              ! (ncol,pver) deep conv. cloud water [kg kg-1]
+    real(kind_phys), intent(in)  :: sh_frac(:, :)               ! (ncol,pver) shallow convective cloud fraction [fraction]
+    real(kind_phys), intent(in)  :: dp_frac(:, :)               ! (ncol,pver) deep convective cloud fraction [fraction]
+    real(kind_phys), intent(in)  :: ast(:, :)                   ! (ncol,pver) physical liquid+ice stratus cloud fraction [fraction]
+    real(kind_phys), intent(in)  :: rei(:, :)                   ! (ncol,pver) ice effective drop size [micron]
```
nusbaume: I might update this comment like so:


```suggestion
    real(kind_phys), intent(in)  :: rei(:, :)                   ! (ncol,pver) ice crystal effective radius [micron]
```

schemes/cloud_water/convective_cloud_water.F90
```diff
@@ -0,0 +1,236 @@
...
+    ! 1. Effective mean in-cloud convective ice/liquid (deep+shallow) !
+    ! 2. Effective mean in-cloud total ice/liquid (ls+convective)     !
+    ! --------------------------------------------------------------- !
+    fresh(:, :) = 0._kind_phys
+    fredp(:, :) = 0._kind_phys
+    frecu(:, :) = 0._kind_phys
+    fretot(:, :) = 0._kind_phys
+
+    do k = 1, pver
+    do i = 1, ncol
```
nusbaume: Nit-pick, but can we indent these loops so that we follow the CAM/SIMA style guide?

schemes/cloud_water/convective_cloud_water.F90
```diff
@@ -0,0 +1,236 @@
...
+
+        cu0_frac = 0._kind_phys
+        cu_icwmr = 0._kind_phys
+
+        ls_frac = ast(i, k)
+        if (ls_frac < frac_limit) then
+          ls_frac = 0._kind_phys
+          ls_icwmr = 0._kind_phys
+        else
+          ls_icwmr = (ls_liq(i, k) + ls_ice(i, k))/max(frac_limit, ls_frac) ! Convert to IC value.
```
nusbaume: I don't think the `max` check is needed here as the if-statement is already ensuring that `ls_frac` is >= to `frac_limit`:


```suggestion
          ls_icwmr = (ls_liq(i, k) + ls_ice(i, k))/ls_frac) ! Convert to IC value.
```

schemes/cloud_water/convective_cloud_water.F90
```diff
@@ -0,0 +1,236 @@
...
+        else
+          kabsi = 0.005_kind_phys + 1._kind_phys/min(max(13._kind_phys, rei(i, k)), 130._kind_phys)
+        end if
+        kabs = kabsl*(1._kind_phys - wrk1) + kabsi*wrk1
+        alpha = -1.66_kind_phys*kabs*pdel(i, k)/gravit*1000.0_kind_phys
+
+        ! Selecting cumulus in-cloud water.
+
+        select case (conv_water_in_rad) ! Type of average
+        case (1) ! Area weighted arithmetic average
```
nusbaume: I wonder if it would be worth having parameters to use here and above for `conv_water_in_rad` instead of just the raw integer numbers?

schemes/cloud_water/convective_cloud_water.F90
```diff
@@ -0,0 +1,236 @@
...
+          cu_icwmr = (sh0_frac*sh_icwmr(i, k) + dp0_frac*dp_icwmr(i, k))/max(frac_limit, cu0_frac)
+        case (2)
+          sh0 = exp(alpha*sh_icwmr(i, k))
+          dp0 = exp(alpha*dp_icwmr(i, k))
+          cu_icwmr = log((sh0_frac*sh0 + dp0_frac*dp0)/max(frac_limit, cu0_frac))
+          cu_icwmr = cu_icwmr/alpha
+        case default ! Area weighted 'arithmetic in emissivity' average.
+!               call endrun ('CONV_WATER_4_RAD: Unknown option for conv_water_in_rad - exiting')
```
nusbaume: Should we set `errmsg` and `errcode` here and return?  I realize it will likely never trigger due to the check above, but it doesn't hurt to be defensive here.

nusbaume: Remove this `case default`?  There is no way for it to trigger given the check above.

schemes/cloud_water/convective_cloud_water.F90
```diff
@@ -0,0 +1,236 @@
...
+        tot0_frac = (ls_frac + cu0_frac)
+
+        select case (conv_water_in_rad) ! Type of average
+        case (1) ! Area weighted 'arithmetic in emissivity' average
+          tot_icwmr = (ls_frac*ls_icwmr + cu0_frac*cu_icwmr)/max(frac_limit, tot0_frac)
+        case (2)
+          tot_icwmr = log((ls_frac*exp(alpha*ls_icwmr) + cu0_frac*exp(alpha*cu_icwmr))/max(frac_limit, tot0_frac))
+          tot_icwmr = tot_icwmr/alpha
+        case default ! Area weighted 'arithmetic in emissivity' average.
+!               call endrun ('CONV_WATER_4_RAD: Unknown option for conv_water_in_rad - exiting')
```
nusbaume: Remove this `case default`?  There is no way for it to trigger given the check above.

schemes/cloud_water/convective_cloud_water.F90
```diff
@@ -0,0 +1,236 @@
...
+
+    ! Local variables:
+    integer         :: i, k                                          ! Lon, lev indices buff stuff.
+    real(kind_phys) :: cu_icwmr                                      ! Convective  water for this grid-box.
+    real(kind_phys) :: ls_icwmr                                      ! Large-scale water for this grid-box.
+    real(kind_phys) :: tot_icwmr                                     ! Large-scale water for this grid-box.
+    real(kind_phys) :: ls_frac                                       ! Large-scale cloud frac for this grid-box.
+    real(kind_phys) :: tot0_frac, cu0_frac, dp0_frac, sh0_frac
+    real(kind_phys) :: kabs, kabsi, alpha, dp0, sh0
+    real(kind_phys) :: wrk1
```
nusbaume: Instead of calling this `wrk1`, I wonder if it would be better to rename it to something like `ls_ice_mass_frac`, or something that helps describe what the variable actually is?

schemes/cloud_water/convective_cloud_water.meta
```diff
@@ -0,0 +1,195 @@
...
+  dimensions = ()
+  intent = in
+[ frac_limit ]
+  standard_name = minimum_cloud_fraction_for_convective_water_in_radiation
+  units = fraction
+  type = real | kind = kind_phys
+  dimensions = ()
+  intent = in
+[ one_mom_clouds ]
+  standard_name = flag_for_one_moment_cloud_microphysics
```
nusbaume: Change this to a verb?  For example:


```suggestion
  standard_name = is_one_moment_cloud_microphysics
```

schemes/cloud_water/convective_cloud_water.meta
```diff
@@ -0,0 +1,195 @@
...
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ totg_ice ]
+  standard_name = gridbox_total_cloud_ice_mixing_ratio_wrt_moist_air_and_condensed_water
+  units = kg kg-1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ conv_liq ]
+  standard_name = in_cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water_due_to_all_convection
```
nusbaume, line 113: I don't think we need the `all` here, as `convection` by itself implies both shallow and deep:


```suggestion
  standard_name = in_cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water_due_to_convection
```

This same request holds for `conv_ice` as well.

schemes/cloud_water/convective_cloud_water.meta
```diff
@@ -0,0 +1,195 @@
...
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ frecu ]
+  standard_name = frequency_of_occurrence_of_convection_with_condensate
+  units = 1
+  type = real | kind = kind_phys
+  dimensions = (horizontal_loop_extent, vertical_layer_dimension)
+  intent = out
+[ fretot ]
+  standard_name = frequency_of_occurrence_of_cloud_with_condensate
```
nusbaume: This standard name feels a little weird to me because by definition a cloud must have condensate (whereas technically convection can occur without it).  Maybe we should say something like "above the minimal level"?


```suggestion
  standard_name = frequency_of_occurrence_of_cloud_with_condensate_above_the_minimal_level
```

schemes/cloud_water/convective_cloud_water_namelist.xml
```diff
@@ -0,0 +1,116 @@
...
+    <group>conv_water_nl</group>
+    <standard_name>control_for_convective_water_in_radiation</standard_name>
+    <units>1</units>
+    <desc>
+      Controls the use of convective in-cloud water in the radiation
+      computation:
+      0 = no convective water (stratiform condensate only),
+      1 = area-weighted arithmetic average of convective and stratiform
+      in-cloud water
+      2 = average in emissivity.
```
nusbaume: I believe option 2 is still area-weighted, so I might change this to something like:


```suggestion
      2 = area-weighted arithmetic average of in-cloud water emissivities
```

schemes/rasch_kristjansson/prognostic_cloud_water.F90
```diff
@@ -75,6 +75,7 @@ subroutine prognostic_cloud_water_init( &
     icritc_in, icritw_in, &
     conke_in, r3lcrit_in, &
     do_psrhmin_in, psrhmin_in, &
+    one_mom_clouds, &
```
nusbaume, line 78: Just a reminder to add this file to the PR description!

schemes/sima_diagnostics/cloud_water_paths_diagnostics.F90
```diff
@@ -0,0 +1,138 @@
+! Diagnostics for cloud_water_paths: the CAM cloud water path history fields.
+! - The in-cloud mixing ratios, water contents and water paths come from
+!   CAM cloud_diagnostics_calc (two-moment branch). The grid-box and
```
nusbaume: Here and below I might remove references to specific CAM subroutines or variables, as this code will (hopefully) outlive CAM, at which point these comments may not be beneficial.

schemes/sima_diagnostics/cloud_water_paths_diagnostics.F90
```diff
@@ -0,0 +1,138 @@
...
+      call history_add_field('ICLDIWP',  'in_cloud_ice_water_path', 'lev', 'avg', 'kg m-2')
+      call history_add_field('ICLDTWP',  'in_cloud_total_water_path', 'lev', 'avg', 'kg m-2')
+      call history_add_field('GCLDLWP',  'gridbox_total_water_path', 'lev', 'avg', 'kg m-2')
+      call history_add_field('TGCLDIWP', 'vertically_integrated_gridbox_ice_water_path', horiz_only, 'avg', 'kg m-2')
+      call history_add_field('TGCLDLWP', 'vertically_integrated_gridbox_liquid_water_path', horiz_only, 'avg', 'kg m-2')
+      call history_add_field('TGCLDCWP', 'vertically_integrated_gridbox_total_water_path', horiz_only, 'avg', 'kg m-2')
+
+      ! History add field calls (CAM micro_pumas_cam field names)
+      call history_add_field('CLDFSNOW', 'liquid_plus_snow_stratiform_cloud_area_fraction', 'lev', 'avg', '1')
+      call history_add_field('CLDFGRAU', 'liquid_plus_graupel_stratiform_cloud_area_fraction', 'lev', 'avg', '1')
```
nusbaume: Can the units not be `fraction` here?

nusbaume (changes requested):
Thanks for bringing in this missed set of schemes @jimmielin!  I have some comments, but most of them are just related to comments or standard names (and like usual please let me know if you have any questions or concerns).  Thanks again!

schemes/cloud_water/cloud_water_paths.F90
jimmielin: updated!

schemes/cloud_water/cloud_water_paths.F90
jimmielin: Thanks, updated!

schemes/cloud_water/cloud_water_paths.F90
jimmielin: Thanks, added `mincld_grau`

schemes/cloud_water/cloud_water_paths.meta
jimmielin: This is pre-existing from RRTMGP 

https://github.com/ESCOMP/atmospheric_physics/blob/4feb6f899eb7f816b1a0d94213e41f3b5114e750/schemes/rrtmgp/rrtmgp_lw_cloud_optics.meta#L74-L76

https://github.com/ESCOMP/atmospheric_physics/blob/4feb6f899eb7f816b1a0d94213e41f3b5114e750/schemes/rrtmgp/rrtmgp_lw_cloud_optics.meta#L80-L82

It's on both CAM-SIMA and atmos_phys across a dozen files so I'm happy to make an issue to sweep the standard names at a later time.

schemes/cloud_water/cloud_water_paths.meta
jimmielin: This is also pre-existing from rrtmgp

https://github.com/ESCOMP/atmospheric_physics/blob/4feb6f899eb7f816b1a0d94213e41f3b5114e750/schemes/rrtmgp/rrtmgp_lw_cloud_optics.meta#L110-L112

schemes/cloud_water/convective_cloud_water.F90
jimmielin: Added, thanks!

schemes/cloud_water/convective_cloud_water.F90
jimmielin: Thanks, updated!

schemes/cloud_water/convective_cloud_water.F90
jimmielin: Indented!

schemes/cloud_water/convective_cloud_water.F90
jimmielin: Thanks, updated!

schemes/cloud_water/convective_cloud_water.F90
jimmielin: Thanks, now module parameters
```fortran
  ! conv_water_in_rad options: how convective in-cloud water enters the radiation inputs
  integer, parameter :: conv_water_none = 0      ! stratiform condensate only
  integer, parameter :: conv_water_arith_avg = 1 ! area-weighted arithmetic average
  integer, parameter :: conv_water_emis_avg = 2  ! area-weighted average in emissivity
```

schemes/cloud_water/convective_cloud_water.F90
jimmielin: Thanks, removed!

schemes/cloud_water/convective_cloud_water.F90
jimmielin: Thanks, removed!

schemes/cloud_water/convective_cloud_water.F90
jimmielin: Thanks, now ls_ice_mass_frac!

schemes/cloud_water/convective_cloud_water.meta
jimmielin: Thanks, updated!

schemes/cloud_water/convective_cloud_water.meta
jimmielin: This is pre-existing, there are other similar names in the approved stdnames spreadsheet with the same suffix, e.g., `detrainment_of_cloud_liquid_water_wrt_moist_air_and_condensed_water_due_to_all_convection`

It is also not possible to rename this easily as that suffix is already used for another meaning. `zm_conv_evap` takes in "general" convection fluxes, shallow and deep respectively, in reevaporation, as the suffix `_due_to_convection`, the interstitials rename `_due_to_shallow_` & `_due_to_deep_` to the general form so they can be used for `zm_conv_evap` shared scheme, then renamed back to `_shallow_`/`_deep_`. So that suffix has a distinct meaning from `_all`

schemes/cloud_water/convective_cloud_water.meta
jimmielin: Agreed. I renamed to `frequency_of_occurrence_of_cloud_with_condensate_above_threshold` to avoid `_level` being confounded with model vertical levels.

schemes/cloud_water/convective_cloud_water_namelist.xml
jimmielin: Thanks, updated!

schemes/sima_diagnostics/cloud_water_paths_diagnostics.F90
jimmielin: Updated!

schemes/sima_diagnostics/cloud_water_paths_diagnostics.F90
jimmielin: Updated to `fraction`!

schemes/rasch_kristjansson/prognostic_cloud_water.F90
jimmielin: Thanks for the reminder - updated here and throughout

schemes/cloud_water/cloud_water_paths.meta
jimmielin: added https://github.com/ESCOMP/atmospheric_physics/issues/442

schemes/cloud_water/cloud_water_paths.meta
jimmielin: added https://github.com/ESCOMP/atmospheric_physics/issues/442

schemes/rasch_kristjansson/rk_stratiform.F90
```diff
@@ -141,11 +135,6 @@ subroutine rk_stratiform_detrain_convective_condensate_run( &
     ! Apply detrainment tendency to cloud liquid water
     tend_cldliq(:ncol,:) = dlf(:ncol,:)
 
-    ! Accumulate precipitation and snow after
-    ! reserved liquid (vertical integral) has now been used
-    ! (snow contribution is zero)
-    prec_str(:ncol) = prec_str(:ncol) - rliq(:ncol)
```
nusbaume, line 147: Claude originally stated that the reason the precipitation clamps in `rk_stratiform_prognostic_cloud_water_tendencies_run` are needed in CAM-SIMA (and CAM) is because of this line being called on the first model time step before precipitation was calculated.  Now that it has been moved to after `rk_stratiform_prognostic_cloud_water_tendencies_run` in the suite, I wonder if it would be worth trying a CAM4 compset in CAM-SIMA with the (un-physical) clamps removed?

schemes/cloud_water/convective_cloud_water.meta
nusbaume: Ahh, right, I forgot about `zm_conv_evap` needing "general" convection values! To be honest I was waffling on this one anyways as `all_convection` isn't wrong, just a little more wordy.  So I'm happy to leave this standard name as-is then.  Thanks for checking!

nusbaume (approved):
Thanks for addressing my requests @jimmielin!  I have one last (optional) request to test a CAM4 compset without the precipitation clamps, although that doesn't have to modify this particular PR.

schemes/rasch_kristjansson/rk_stratiform.F90
jimmielin: Thanks! I launched the run without the clamp on Izumi and will compare the results and report back.

schemes/cloud_water/cloud_water_paths.F90
```diff
@@ -0,0 +1,174 @@
...
+    icgrauwp = 0._kind_phys
+    cldfgrau = 0._kind_phys
+
+    do k = top_lev, pver
+      do i = 1, ncol
+        ! ------------------------------
+        ! Adjust cloud fraction for snow
+        ! ------------------------------
+        cldfsnow(i, k) = cld(i, k)
+        ! If cloud and only ice (no convective cloud or ice), then set to 0.
```
peverwhee: Not your error, but this comment (and the one on line 156) is confusing to me. Is this more accurate? If not, is there a different wording that make it seem like we're checking to see if "ice" is present but also that "ice" isn't present?

```suggestion
        ! If cloud and only ice (no convective cloud or liquid), then set to 0.
```

peverwhee: also, would it be more accurate to say "if cloud and only **snow**"?

peverwhee (approved):
one wee comment gripe, but happy to give my approval!

schemes/rasch_kristjansson/rk_stratiform.F90
jimmielin: Hi @nusbaume - Claude compared the with-floor and without-floor runs and concluded they're different at roundoff level starting from step ~16 (then the error propagates due to chaos). I think perhaps we could just get rid of the un-physical clamp, since the model runs stable now for 5 years on FHIST_C4 without it: please let me know! I'm happy to remove the clamp in this PR together with the rest of the RK changes.

nusbaume: Great, thanks for checking @jimmielin!  I agree that removing the clamp would be the best choice scientifically, so if the model runs stably without it then I say let's get rid of it.  Thanks again!

Issue #435, opened 2026-08-19
Move constituent registration out of rrtmgp_constituents to schemes that own/provide the constituents


jimmielin:
Right now `rrtmgp_constituents_register` registers a dynamic constituent for every entry in the `rad_climate` namelist, e.g.,

https://github.com/ESCOMP/atmospheric_physics/blob/64b254f0de4780e6f645997e9c4381280738aaef/schemes/rrtmgp/rrtmgp_constituents.F90#L75-L80

But rrtmgp_constituents does not *provide* values for any of those constituents; other schemes do, e.g., prescribe_radiative_gas_concentrations.

We should move the registration out of the `rrtmgp_constituents` to be consistent with our approach that schemes that own/provide the constituent registers it.

Additionally:

CAM's `rad_climate` entries are `source:source_name:gas_name`, e.g.
`'N:CFC11STAR:CFC11'`: the field named CFC11STAR (CFC11 with the minor
halon species folded in for LW radiation) feeds radiation's CFC11 gas slot.
The name on the left of the mapping identifies where the data comes from and
the name on the right identifies which radiation gas it fills.

The CAM-SIMA format dropped this, entries are `flag:long_name:gas_name`,
the third field is used as both the registered constituent's standard name
and the radiation gas name, and `rrtmgp_constituents_run` looks
constituents up by gas name. There is no way to say "constituent X feeds
radiation gas Y", so surrogates like CFC11STAR cannot be expressed. CAM7's
ghg_mam4 configuration uses `N:CFC11STAR:CFC11`, so this is needed for CAM7.

I think we could change the middle field to be the `diag_name` (which was introduced after RRTMGP was CCPPized)

tagging @peverwhee @nusbaume !

jimmielin:
I will work this as part of FHIST_C4 work so I've self-assigned it, since I need to accommodate for LBCs writing into the constituents as well.

PR #437, opened 2026-08-23
Mods needed to run CAM4 F-compset (AMIP) configurations

nusbaume:
Tag name (The PR title should also include the tag name): atmos_phys0_28_001
Originator(s): nusbaume  (Claude Opus 5 helped find the issues).

Added some code modifications needed for the CAM4 + BAM physics/chemistry suite to run stably in an F-compset (i.e. AMIP or prescribed SSTs/sea ice) configuration.

M       schemes/rrtmgp/rrtmgp_constituents_namelist.xml
 - Removed aquaplanet flag when setting radiatively active gases (which should eventually be managed by the chemistry scheme anyways, see #435).

 - Add fixer for negative surface precipitation, which matches what is done in CAM.

Yes, the RK precipitation adjustment results in answer differences.

I did a five year run and looked at the results.

### DIFF
--- a/schemes/rasch_kristjansson/rk_stratiform.F90
+++ b/schemes/rasch_kristjansson/rk_stratiform.F90
@@ -5,7 +5,6 @@ module rk_stratiform
 
   implicit none
   private
-  save
 
   ! public CCPP-compliant subroutines
   !   note: cloud_fraction_perturbation_run calls the compute_cloud_fraction
@@ -459,6 +458,15 @@ contains
 
    prec_str(:ncol) = prec_str(:ncol) + prec_pcw(:ncol)
    snow_str(:ncol) = snow_str(:ncol) + snow_pcw(:ncol)
+
+   ! RK can occassionally produce negative precipitation fluxes, so force
+   ! it to be greater than or equal to zero.  Also ensure that snow is not
+   ! greater than the total precipitation. This is a non-physical adjustment,
+   ! but is the same technique that the original CAM4 model used:
+   do i = 1, ncol
+      if (prec_str(i) < 0._kind_phys) prec_str(i) = 0._kind_phys
+      if (snow_str(i) > prec_str(i))  snow_str(i) = prec_str(i)
+   end do
 
   end subroutine rk_stratiform_prognostic_cloud_water_tendencies_run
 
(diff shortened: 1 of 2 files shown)


schemes/rasch_kristjansson/rk_stratiform.F90
```diff
@@ -460,6 +459,15 @@ subroutine rk_stratiform_prognostic_cloud_water_tendencies_run( &
...
 
+   ! RK can occassionally produce negative precipitation fluxes, so force
+   ! it to be greater than or equal to zero.  Also ensure that snow is not
+   ! greater than the total precipitation. This is a non-physical adjustment,
+   ! but is the same technique that the original CAM4 model used:
+   do i = 1, ncol
+      if (prec_str(i) < 0._kind_phys) prec_str(i) = 0._kind_phys
+      if (snow_str(i) > prec_str(i))  snow_str(i) = prec_str(i)
+   end do
+
```
jimmielin, line 470: Since this fix is going into `rk_stratiform_prognostic_cloud_water_tendencies_run` (which is called by CAM `rk_stratiform_cam.F90`) maybe we need to remove the potentially now-duplicate (although idempotent) check in the CAM source code? Perhaps a PR is already forthcoming, in which case please feel free to ignore this comment!

nusbaume: Thanks for the comment @jimmielin!  However until I know that the other microphysics schemes (e.g. PUMAS) don't also need the fix I want to hold off on removing the check in CAM, as it could break things.

Also, just an FYI that I want to leave the adjuster with the problematic schemes in atmospheric_physics so that we know which schemes may need to be fixed in the future (and it helps keep science code out of CAM-SIMA itself).

jimmielin: Ah I see, so the adjuster is not on the RK end (i.e., it's somewhere for all microphysics on the CAM side?), sorry I missed it during the initial CCPPization! I agree we want to keep it here for now, or potentially in a dedicated adjuster scheme if it ends up getting repeated too often.

jimmielin (approved):

peverwhee (approved):

PR #444, opened 2026-09-08
Various fixes for enablement of FHIST_C4 (CAM4 transient compsets); CCPPize mo_flbc

### LINKED ISSUE #435: Move constituent registration out of rrtmgp_constituents to schemes that own/provide the constituents

jimmielin:
Originator(s): @jimmielin coauthored with `claude-fable:5`, `claude-fable:5.1`, `claude-opus:5`

- Various mods for FHIST_C4 enablement, incl.
- Closes https://github.com/ESCOMP/atmospheric_physics/issues/435
- CCPPize `mo_flbc` -> [prescribe_lower_boundary_conditions.F90](https://github.com/ESCOMP/atmospheric_physics/pull/444/files#diff-4350e70ba82c275ba2b2201a4b5c11f13867c6259943b72fd61c11e6431fc940) for transient GHG boundary conditions and for future ghg_mam4
- Add transient defaults to various prescribed schemes
- Wire up RK to conv_water and radiation by providing MG-style optical inputs [rk_stratiform_mg_optics_inputs.F90](https://github.com/ESCOMP/atmospheric_physics/pull/444/files#diff-15f9ec90b7a7bffc1dc4b6dc6aee5246939f8a22b3206f35099a917da1785b97)
- Port slingo/ebert-curry liq/ice optics option in addition to existing mitchell/gammadist option
- Add cloud cover and precipitation diagnostics schemes
- Add coupler surface fluxes to sima_diagnostics scheme


schemes/chemistry/prescribed_aerosol_deposition_flux_namelist.xml
```diff
@@ -102,6 +102,7 @@
       <value>aerosoldep_monthly_1849-2006_1.9x2.5_c090803.nc</value>
       <!-- pre-industrial (1850) climatology -->
       <value sim_year="1850">aerosoldep_monthly_1850_mean_1.9x2.5_c090421.nc</value>
+      <value sim_year="1850-2015">aerosoldep_monthly_1849-2006_1.9x2.5_c090803.nc</value>
```
nusbaume: I know CAM does this, but I don't think we want to use an actual date range for the `sim_year` flag, as that can change with new input datasets and thus could become misleading.  I wonder if instead it might be better to have the default settings be for historical configurations, and then just set different values if `sim_year` is provided (which should only happen for constant year compsets)?

schemes/radiation_utils/solar_irradiance_data.F90
```diff
@@ -26,6 +26,10 @@ module solar_irradiance_data
   real(kind_phys), allocatable :: irrad_fac(:)
   real(kind_phys), allocatable :: etf_fac(:)
   real(kind_phys), allocatable :: lambda(:)
+  ! Bracketing time slices read from the file, interpolated every timestep;
+  ! module state because timesteps between reads reuse the previous slices
```
nusbaume: I would remove the second comment line here, as I think it's relatively obvious why it is at the module-level.  I might also make it clear that it is irradiance data:

```suggestion
  ! Bracketing time slices of irradiance read from the file, interpolated every timestep.
```

schemes/radiation_utils/solar_irradiance_data.F90
```diff
@@ -191,8 +195,9 @@ subroutine solar_irradiance_data_init(irrad_file_path, solar_data_type, solar_da
        return
     end if
 
-    ! Check what the file contains
-    call file_reader%get_var('ssi', ssi, errmsg, errflg)
+    ! Check what the file contains.
+    ! Use a single-element read instead of reading the full variable every PET:
+    call file_reader%get_var('ssi', ssi, errmsg, errflg, (/1, 1/), (/1, 1/))
```
nusbaume: I am never sure how common the "PET" terminology is, so I would just say "task/thread".  I also would use the more modern Fortran array syntax:

```suggestion
    ! Use a single-element read instead of reading the full variable on every task/thread:
    call file_reader%get_var('ssi', ssi, errmsg, errflg, [1, 1], [1, 1])
```

In fact it might be good to use that array syntax throughout the whole physics scheme?

test/test_schemes/ozone_constituent_stub.F90
```diff
@@ -0,0 +1,56 @@
...
+contains
+
+!> \section arg_table_ozone_constituent_stub_register  Argument Table
+!! \htmlinclude ozone_constituent_stub_register.html
+  subroutine ozone_constituent_stub_register(constituents, errmsg, errcode)
+    use ccpp_constituent_prop_mod, only: ccpp_constituent_properties_t
+    use ccpp_chem_utils,           only: chem_constituent_qmin
+
+    type(ccpp_constituent_properties_t), allocatable, intent(out) :: constituents(:)
+    character(len=512), intent(out) :: errmsg
```
nusbaume: This should be assumed-length:

```suggestion
    character(len=*), intent(out) :: errmsg
```

This same request holds for the other `errmsg` output variables as well in this PR.

schemes/sima_diagnostics/rrtmgp_lw_diagnostics.F90
```diff
@@ -112,8 +113,13 @@ subroutine rrtmgp_lw_diagnostics_run(num_diag_subcycles, icall, active_calls, fl
...
 
-      ! Don't do anything if this subcycle is inactive or we're not configured to write radiation output
-      if ((.not. active_calls(diag_index+1)) .or. (.not. write_output)) then
+      ! Don't do anything if this subcycle is inactive,
+      !  we're not configured to write radiation output, or
+      !  radiation was not run this timestep.
+      !
+      ! The flux objects are zeroed on non-radiation timesteps so they should not be
+      ! written to average accumulation on non-radiation timesteps.
+      if ((.not. dolw) .or. (.not. active_calls(diag_index+1)) .or. (.not. write_output)) then
```
nusbaume: I think this is fine for now, but what happens if instantaneous output has been selected?  Will it just use the previously set value, automatically set it to zero, or something else?

Also I might update the comment slightly:

```suggestion
      ! Don't do anything if this subcycle is inactive,
      !  we're not configured to write radiation output, or
      !  radiation was not run this timestep.
      !
      ! The flux objects are zeroed on non-radiation timesteps so they should not be
      ! written to accumulated history fields on non-radiation timesteps.
      if ((.not. dolw) .or. (.not. active_calls(diag_index+1)) .or. (.not. write_output)) then
```

schemes/sima_diagnostics/precipitation_diagnostics.F90
```diff
@@ -0,0 +1,74 @@
+! Surface precipitation rate diagnostics: PRECC/PRECL/PRECT
+!
+! Uses the same per-process rates sent to the coupler by set_surface_coupling_vars
+! to match what the surface components receives.
+module precipitation_diagnostics
```
nusbaume: Instead of having the surface fluxes be in `sima_state_diagnostics`, and the surface precipitation diagnostics here, would it make more sense to combine them both into a new `surface_coupling_diagnostics`, or something that makes it clear that without `set_surface_coupling_vars` these diagnostics don't mean much for the physical climate?  It would also help avoid the situation where a person tries to output, say `SHFLUX` in an idealized physics setup where it is always zero (and thus just wasting disk space).

Of course if you have a different opinion, or would like to discuss it more, then just let me know.  Thanks!

schemes/sima_diagnostics/cloud_cover_diagnostics.F90
```diff
@@ -0,0 +1,162 @@
+! Cloud cover diagnostics
+! total/low/mid/high cloud cover assuming maximum-random overlap and 3D cloud fraction.
+!
+! Based on cldsav, cloud_cover_diags from CAM: W. Collins
```
nusbaume: I might get rid of this line, as someday CAM will (hopefully) be gone:

```suggestion
```

schemes/sima_diagnostics/cloud_cover_diagnostics.F90
```diff
@@ -0,0 +1,162 @@
...
+   public :: cloud_cover_diagnostics_init
+   public :: cloud_cover_diagnostics_run
+
+   ! Pressure bounds of the low/mid/high cloud cover ranges [Pa]
+   real(kind_phys), parameter :: plowmax = 120000._kind_phys ! Max prs for low cloud cover range
+   real(kind_phys), parameter :: plowmin =  70000._kind_phys ! Min prs for low cloud cover range
+   real(kind_phys), parameter :: pmedmax =  70000._kind_phys ! Max prs for mid cloud cover range
+   real(kind_phys), parameter :: pmedmin =  40000._kind_phys ! Min prs for mid cloud cover range
+   real(kind_phys), parameter :: phghmax =  40000._kind_phys ! Max prs for hgh cloud cover range
+   real(kind_phys), parameter :: phghmin =   5000._kind_phys ! Min prs for hgh cloud cover range
```
nusbaume: Total nit-pick, but can we spell out `pressure` in the comments here?

```suggestion
   real(kind_phys), parameter :: plowmax = 120000._kind_phys ! Max pressure for low cloud cover range
   real(kind_phys), parameter :: plowmin =  70000._kind_phys ! Min pressure for low cloud cover range
   real(kind_phys), parameter :: pmedmax =  70000._kind_phys ! Max pressure for mid cloud cover range
   real(kind_phys), parameter :: pmedmin =  40000._kind_phys ! Min pressure for mid cloud cover range
   real(kind_phys), parameter :: phghmax =  40000._kind_phys ! Max pressure for hgh cloud cover range
   real(kind_phys), parameter :: phghmin =   5000._kind_phys ! Min pressure for hgh cloud cover range
```

schemes/rrtmgp/utils/rrtmgp_cloud_optics_setup.F90
```diff
@@ -52,6 +60,20 @@ subroutine rrtmgp_cloud_optics_setup_init(liq_filename, ice_filename, errmsg, er
...
+    liq_cld_optics = liq_cld_optics_nl
+    ice_cld_optics = ice_cld_optics_nl
+    if (trim(liq_cld_optics) /= 'gammadist' .and. trim(liq_cld_optics) /= 'slingo') then
+       write(errmsg,'(a,a,a)') sub, ': liq_cld_optics must be either slingo or gammadist, got ', trim(liq_cld_optics)
+       errflg = 1
+       return
+    end if
+    if (trim(ice_cld_optics) /= 'mitchell' .and. trim(ice_cld_optics) /= 'ebertcurry') then
+       write(errmsg,'(a,a,a)') sub, ': ice_cld_optics must be either ebertcurry or mitchell, got ', trim(ice_cld_optics)
+       errflg = 1
```
nusbaume: I would put quotes around the bad optics entries, just in case:

```suggestion
    if (trim(liq_cld_optics) /= 'gammadist' .and. trim(liq_cld_optics) /= 'slingo') then
       write(errmsg,'(a,a,a)') sub, ': liq_cld_optics must be either slingo or gammadist, got "', trim(liq_cld_optics), '"'
       errflg = 1
       return
    end if
    if (trim(ice_cld_optics) /= 'mitchell' .and. trim(ice_cld_optics) /= 'ebertcurry') then
       write(errmsg,'(a,a,a)') sub, ': ice_cld_optics must be either ebertcurry or mitchell, got "', trim(ice_cld_optics), '"'
       errflg = 1
```

suites/suite_cam4.xml
```diff
@@ -17,14 +17,22 @@
...
     <!-- INITIALIZE CLOUD FRACTION NAMELIST (init phase only) -->
     <scheme>cloud_fraction_options</scheme>
 
+    <!-- Prescribe fixed GHG concentrations -->
+    <scheme>prescribe_radiative_gas_concentrations</scheme>
+
+    <!-- Prescribe time-varying lower boundary concentrations
+         only active when flbc_file is set.
+         Overrides the GHG values for the species in flbc_list over
+         prescribe_radiative_gas_concentrations -->
```
nusbaume: I might simplify this comment some:

```suggestion
    <!-- Prescribe time-varying lower boundary concentrations.
         Note that this can overwrite previously-set constituents 
         like GHGs -->
```

suites/suite_cam4.xml
```diff
@@ -242,6 +251,23 @@
     <!-- State diagnostics -->
     <scheme>sima_state_diagnostics</scheme>
 
+    <!--
+      Cloud water inputs for radiation (in-cloud water paths).
+      For parity with CAM4, conv_water_in_rad = 0 option is used;
+      convective_cloud_water provides outputs as zeros
+      and cloud_water_paths uses stratiform condensate only.
+      qsout/qgout are host-registry zeros since RK has no prognostic snow/graupel.
+    -->
```
nusbaume: I would clean-up this comment, as I feel like it has more info then I would want in the SDF (especially since a user could change `conv_water_in_rad` if they wanted to).

```suggestion
    <!-- Cloud water inputs for radiation (in-cloud water paths) -->
```

suites/suite_cam4.xml
```diff
@@ -242,6 +251,23 @@
...
+    <scheme>convective_cloud_water</scheme>
+    <scheme>convective_cloud_water_diagnostics</scheme>
+    <scheme>cloud_water_paths</scheme>
+    <scheme>cloud_water_paths_diagnostics</scheme>
+
+    <!-- Derive MG-convention optics inputs (dei, pgam, lamc) from the RK
+         effective radii so the gammadist liquid / Mitchell ice RRTMGP
+         cloud optics see the RK clouds. -->
```
nusbaume: I would update/simplify this comment as well:

```suggestion
    <!-- Derive PUMAS-convention optics inputs (dei, pgam, lamc) from the RK
         effective radii -->
```

schemes/rasch_kristjansson/rk_stratiform_mg_optics_inputs.F90
```diff
@@ -0,0 +1,84 @@
+! Derive the MG-convention cloud optics inputs (dei, pgam, lamc)
```
nusbaume: I think I might change the scheme name to `rk_stratiform_pumas_optics_inputs` (as PUMAS has fully replaced MG), and replace references to "MG" with "PUMAS" in the comments here and below.

schemes/rasch_kristjansson/rk_stratiform_mg_optics_inputs.F90
```diff
@@ -0,0 +1,84 @@
...
+! for RRTMGP gammadist liquid / Mitchell ice cloud optics
+! from the RK (CAM4 cldefr) climatological effective radii rel/rei.
+!
+! RK microphysics carries no size distribution information, so the gamma
+! distribution shape parameter pgam is fixed and the slope lamc is chosen to
+! reproduce the RK liquid effective radius
+!                           re = (pgam+3)/(2*lamc)
+! Points without condensate are left at zero which are skipped by the optics/
```
nusbaume: Remove the trailing `/` (and slightly change grammar)?

```suggestion
! Points without condensate are left at zero and are skipped by the optics.
```

